PR #3, opened 2016-08-17
Speed up casting by not killing apps known to be nonblocking

theychx:
Greetings! Just a little improvement to make catt faster

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -66,8 +66,11 @@ class CastController:
         time.sleep(0.2)
 
     def play_media(self, url, content_type="video/mp4"):
-        self.kill()
-        time.sleep(5)
+        NON_BLOCKING_APPS = ['Backdrop', 'Default Media Receiver',
+                             'BubbleUPnP']
+        if self.cast.app_display_name not in NON_BLOCKING_APPS:
+            self.kill()
+            time.sleep(5)
         self.cast.play_media(url, content_type)
 
     def play(self):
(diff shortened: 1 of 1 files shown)


catt/controllers.py
```diff
@@ -66,8 +66,12 @@ def __init__(self):
         time.sleep(0.2)
 
     def play_media(self, url, content_type="video/mp4"):
-        self.kill()
-        time.sleep(5)
+        NON_BLOCKING_APPS = ['Backdrop', 'Default Media Receiver',
+                             'BubbleUPnP']
+        if self.cast.app_display_name not in NON_BLOCKING_APPS:
+            self.kill()
+            while self.cast.app_display_name != 'Backdrop':
```
skorokithakis: You're sure this updates every call, right?

skorokithakis:
Hello! Can you explain the rationale behind this a bit, and also how it works?

theychx:
> You're sure this updates every call, right?
Yep, tested it with serveral apps, it always gets the id right.

The point is that only a genuine "Custom Receiver" app (such as YouTube) might  exhibit the blocking behaviour. Theres no need to kill the currently running app and wait 5 seconds, if this app will let go of control voluntarily.
In the case of a  "Custom Receiver" app running, I changed the code so instead of just waiting 5 seconds, we poll for "Backdrop" to appear, and thus save some time here as well.

PS: I added BubbleUpnp to the list of nonblocking apps because it is an iteration of the "Styled Media Receiver" which I believe behaves just as the "Default Media Receiver"

skorokithakis:
Fantastic, thank you very much for the contribution!

PR #4, opened 2016-08-18
Add additional volume and info methods

theychx:
Howdy. A few further improvements here: "volumeup" and "volumedown" for incremental volume change, useful for keyboard-shortcuts and the like. And "info" for a complete dump of the contents of  cast.media_controller.status, which I think would be useful for testing/debugging purposes.


skorokithakis:
Fantastic, thank you!

PR #7, opened 2016-12-01
Remove unnecessary code from play_media

theychx:
Howdy. A while back I gutted play_media of the checks we are doing now, to see if they were actually making a difference. It turns out they are not. I have had no failure in this period with this gutted version of play_media. I am pretty sure that, as long as we exclusively are using the 'Default Media Receiver' for playback, no blocking could occur.


skorokithakis:
Fantastic, thank you!

PR #8, opened 2016-12-02
Allow IP or chromecast name to be provided

pschmitt:
This adds the options `--ip` and `--name` to bypass auto-detection and the cache. This is particularly useful if numerous chromecast devices are present on the network.

A few issues remain in this PR:
1. `--ip` and `--name` should be mutually exclusive. I'm not quite sure how to define them as such using click.
2. The use of global variables `chromecast_ip` and `chromecast_name` is not quite sexy

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -10,6 +10,9 @@ from .controllers import get_stream_info, CastController, Cache
 from .http_server import serve_file
 
 
+chromecast_ip = None
+chromecast_name = None
+
 def get_local_ip(cc_host):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     s.connect((cc_host, 0))
@@ -20,15 +23,24 @@ def get_local_ip(cc_host):
 @click.option("--delete-cache", is_flag=True, help="Empty the Chromecast "
               "discovery cache. Specify this if you're having errors connecting to "
               "the Chromecast.")
-def cli(delete_cache):
+# FIXME --ip and --name should be mutually exclusive
+@click.option("--ip", help="IP Address of the target Chromecast device")
+@click.option("--name", help="Name of the target Chromecast device")
+def cli(delete_cache, ip, name):
     if delete_cache:
         Cache().clear()
+    if ip:
+        global chromecast_ip
+        chromecast_ip = ip
+    elif name:
+        global chromecast_name
+        chromecast_name = name
 
 
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url")
 def cast(video_url):
-    cast = CastController()
+    cast = CastController(chromecast_ip, chromecast_name)
     cc_name = cast.cast.device.friendly_name
 
     if "://" not in video_url:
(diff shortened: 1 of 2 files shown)


skorokithakis:
Thank you for the PR! Since this is a non-trivial change, I will review it soon in detail, but this functionality is great, thanks again.

skorokithakis:
I haven't managed to review both of these properly, but #9 seems a bit more fleshed out than #8 at a first glance. I would like to only merge one, since they're overlapping, so I'd like you guys to tell me what you think. @pschmitt, @theychx?

theychx:
In my opinion, it doesn't really make sense to give users the ability to target specific devices, without adapting the cache to this. And with the new cache, users can target devices without having to fiddle with ip-adresses AND without catt doing auto-discovery.

skorokithakis:
@theychx I would agree with taht, it's better if the new devices can be cached on a per-device basis as well.

pschmitt:
Yeah, #9 is definitely cleaner. Being able to set the IP address of the target device would be a nice feature though. I'll close this one.

skorokithakis:
I agree that it would be nice to be able to set the IP address. @theychx, would that be easy to add to your PR?

skorokithakis:
Thank you for your help, @pschmitt!

pschmitt:
A check on the cli provided name using [ipaddress](https://docs.python.org/3/library/ipaddress.html) should be enough. If the provided "name" is an IP address use it, else use the default logic.

theychx:
@pschmitt
@skorokithakis, yes that would be possible,  although I do not really understand why that would be useful.

skorokithakis:
Hmm, true, I guess the cache would take care of the IP on the first run, wouldn't it... @pschmitt, can you tell us a bit more on what the use case for setting the IP was?

pschmitt:
It is quite simple and boring: I may know the CC IP address but not its name.

skorokithakis:
Ah, okay. I would like it if the `--name` option could be used with an IP as well, but it's hard to tell what an IP looks like (might be IPv4, IPv6, decimal or even a hostname), so I'm ambivalent about adding an `--ip` option, I don't know if it will be useful to enough people to justify the added complexity...

theychx:
I have basic config file read/write working, and with that implementation it would not be feasible to be able to pass ip's and names to the same option. So if we end up deciding on the user being able to select cc via ip, it would need to be as a seperate option (which would not be a problem).

pschmitt:
+1 for the separate option. The user should know if he wants to address his CC device via an IP address/hostname.

theychx:
I have no idea on how to make those options mutually exclusive tho. But the way I would do it, ip's would take precedence over names. It's just suboptimal.

theychx:
> I have no idea on how to make those options mutually exclusive tho
Ah, I just figured out a way to do just that. A bit hackish, but it would work, so no problem.

theychx:
@skorokithakis, I changed my mind on this, as I think it would be useful for testing.

Issue #5, opened 2016-08-21
Catt does not support multiple casting devices


theychx:
Hello. So I just bought a Chromecast Audio to go along with my CC2.As you know, this breaks my current setup where I use catt in several ways. So I am proposing the following:

Catt should support an option to set `-d "CCaudio"` or `-d "192.168.1.200"` to make commands specific to this device. Catt should also employ a rudimentary config file (youtube-dl style) so that the user can set a default device (or perhaps other future flags). This could go in `~/.config/catt/config` or perhaps just `~/.catt`. I am not shure how the current caching scheme fits into this, but it would certainly have to change.

Would you consider merging a PR implementing these ideas?

skorokithakis:
Yes, definitely. This is very important, and the only reason I haven't implemented it is because I don't have a second Chromecast. As for the caching, the cache key will just have to include the device name.

theychx:
Cool. I have some further thoughts about the cache. Consider this scenario:

I have setup my CC2 as my default device in `~/.catt`
Occasionally I would like to use catt with my CC Audio like this:
`catt -d "CCaudio" https://www.youtube.com/watch?v=SdHBkYoUXOs`
Now whenever I  want to control playback on the  CC Audio, I would have to do this:
`catt -d "CCaudio" seek 300`

Maybe instead of changing the format of the cache itself, we could change the caching mechanism so that we write to cache everytime a cast command is run. This way you could manipulate the active CC with a command like this: `catt -c seek 300`. Or is this too involved?

skorokithakis:
I think that might be getting a bit too involved, in that it might lead to surprising behaviour because the state is no longer in the command-line options, which is probably not a good idea. Also, given that your terminal can easily autocomplete to previous commands, and that you can set the default in the preferences, I think I'm leaning towards disliking this.

However, there's no reason why there couldn't be a device alias in the preferences file, so you could do -d 1, for example.

theychx:
Good points. I'll keep you posted

skorokithakis:
I'm going to reopen this. Trying to use these features now, I realize that `--write-config` shouldn't be a flag, it should be a full-blown command (`catt write-config`). I'm going to reimplement it.

theychx:
I actually implemented it like this at first, but changed my mind, as the flag seemed like the natural choice. Maybe I was wrong, as the command perhaps is more robust. What convinced you?

skorokithakis:
[...]

PR #9, opened 2016-12-02
feat: Add user selectable CC device

theychx:
Hi. So here's part 1 of the solution to https://github.com/skorokithakis/catt/issues/5 (config still to come).

I am not very good friends with click, so I hope my implementation is somewhat correct.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -30,10 +30,6 @@ class ChromecastDeviceError(ClickException):
 class Cache:
     def __init__(self, duration=3 * 24 * 3600,
                  cache_dir=os.path.join(tempfile.gettempdir(), "catt_cache/")):
-        def _initialize_cache(filename):
-            with open(filename, "w") as cache:
-                json.dump({}, cache)
-
         self.cache_dir = cache_dir
         try:
             os.mkdir(cache_dir)
@@ -45,9 +41,16 @@ class Cache:
 
         if os.path.exists(cache_filename):
             if os.path.getctime(cache_filename) + duration < time.time():
-                _initialize_cache(cache_filename)
+                self._initialize_cache()
         else:
-            _initialize_cache(cache_filename)
+            self._initialize_cache()
+
+    def _initialize_cache(self):
+        data = {}
+        devices = pychromecast.get_chromecasts()
+        for device in devices:
+            data[device.name] = device.host
+        self._write_cache(data)
 
     def _read_cache(self):
         with open(self.cache_filename, "r") as cache:
@@ -59,16 +62,14 @@ class Cache:
 
     def get(self, name):
         data = self._read_cache()
-        if not data:
-            return
         if not name:
             devices = list(data.keys())
             devices.sort()
-            return data[devices[0]]
+            return (devices[0], data[devices[0]])
         try:
-            return data[name]
+            return (name, data[name])
         except KeyError:
-            pass
+            return (name, None)
 
     def set(self, name, value):
         data = self._read_cache()
@@ -84,22 +85,15 @@ class Cache:
 
 class CastController:
     def __init__(self, device_name):
-        def _check_device(device):
-            if not device:
-                raise ChromecastDeviceError("Device not found.")
-
         cache = Cache()
-        cached_chromecast = cache.get(device_name)
+        cached_name, cached_ip = cache.get(device_name)
 
         try:
-            self.cast = pychromecast.Chromecast(cached_chromecast)
+            self.cast = pychromecast.Chromecast(cached_ip)
         except pychromecast.error.ChromecastConnectionError:
-            if device_name:
-                self.cast = pychromecast.get_chromecast(friendly_name=device_name)
-                _check_device(self.cast)
-            else:
-                self.cast = pychromecast.get_chromecast()
-                _check_device(self.cast)
+            self.cast = pychromecast.get_chromecast(friendly_name=cached_name)
+            if not self.cast:
+                raise ChromecastDeviceError("Device not found.")
             cache.set(self.cast.name, self.cast.host)
 
         time.sleep(0.2)
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -67,7 +67,7 @@ class Cache:
         if not name:
             devices = list(data.keys())
             devices.sort()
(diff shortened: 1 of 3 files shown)


skorokithakis:
Thank you for the PR! Isn't there quite a bit of overlap with #8, though?

theychx:
> Thank you for the PR! Isn't there quite a bit of overlap with #8, though?

Yeah, I know. I didn't realize before I had pushed the first commit.

theychx:
As we are now doing auto-discovery during cache initialization, current test-suite will only pass if there are no cc's on network.

theychx:
@skorokithakis, I changed my mind on this, as I think it would be useful for testing.

skorokithakis:
Is this ready to review?

theychx:
Yep

skorokithakis:
Great, thank you. Will review ASAP.

catt/controllers.py
```diff
@@ -22,29 +23,60 @@ def get_stream_info(video_url):
...
+        with open(self.cache_filename, "r") as cache:
+            return json.load(cache)
+
+    def _write_cache(self, data):
+        with open(self.cache_filename, "w") as cache:
+            json.dump(data, cache)
+
+    def get(self, name):
+        data = self._read_cache()
+        if not data:
```
skorokithakis: Can you add one or two comments here? They would make the intent of this code easier to follow.

catt/controllers.py
```diff
@@ -54,15 +86,23 @@ def clear(self):
 
 
 class CastController:
-    def __init__(self):
+    def __init__(self, device_name):
         cache = Cache()
+        cached_ip = cache.get(device_name)
```
skorokithakis: Does this have to be the exact name? It would be nice if we could do a substring match "if name in chromecast_name".

catt/controllers.py
```diff
@@ -54,15 +86,23 @@ def clear(self):
...
+        try:
+            if not cached_ip:
+                raise ValueError
+            self.cast = pychromecast.Chromecast(cached_ip)
+        except (pychromecast.error.ChromecastConnectionError, ValueError):
+            if device_name:
+                self.cast = pychromecast.get_chromecast(friendly_name=device_name)
+            else:
+                self.cast = pychromecast.get_chromecast()
+            if not self.cast:
```
skorokithakis: The following can probably be outside the `except`.

catt/controllers.py
```diff
@@ -22,29 +23,60 @@ def get_stream_info(video_url):
...
+            self._initialize_cache()
+
+    def _initialize_cache(self):
+        data = {}
+        devices = pychromecast.get_chromecasts()
+        for device in devices:
+            data[device.name] = device.host
+        self._write_cache(data)
+
+    def _read_cache(self):
```
skorokithakis: We can probably store the loaded/written data in this instance in the future, but for now we're only calling `.get` once, so it's fine.

catt/controllers.py
skorokithakis: I mean, if my Chromecast is called "Stavros' Chromecast", can I specify the name as "stavros" and still have it match?

catt/controllers.py
theychx: True

catt/controllers.py
skorokithakis: Ah, yes, sorry. I better go to bed and resume the review when it's not 6 am :/

catt/controllers.py
```diff
@@ -22,29 +23,60 @@ def get_stream_info(video_url):
...
+                 cache_dir=os.path.join(tempfile.gettempdir(), "catt_cache")):
         self.cache_dir = cache_dir
         try:
             os.mkdir(cache_dir)
         except:
             pass
 
-    def _get_cache_filename(self, key):
-        return os.path.join(self.cache_dir, key)
+        cache_filename = os.path.join(cache_dir, "chromecast_hosts")
```
theychx: I was a bit unsure about the finer details in Cache.__init__. What would be your preference?

catt/controllers.py
theychx: That sounds like a rabbit hole to me. 'Living Room' 'Bed Room' etc....

catt/controllers.py
theychx: ` if not data` is for the scenario where cache has been initialized with no cc's on the network (empty dict).
`if not name` is an attempt to consistently return the same ip when the user does not specify a device, hence the alphabetical sort (not reliable if the user adds/removes cc's from the network in between commands).
Let me now if want the code itself commented, I wasn't sure what you meant.

catt/controllers.py
skorokithakis: I was using a function for no particular reason, but since you're directly assigning, might as well just do:

~~~
self.cache_filename = os.path.join(cache_dir, "chromecast_hosts")
~~~

I think that's quite clear and works well.

catt/controllers.py
skorokithakis: Sorry, I meant comments in the code. Generally, if I'm reading something and thinking "I wonder why they're doing this here", it means the code needs a few comments explaining the intent.

catt/controllers.py
skorokithakis: Yes, but then you would only specify "living" or "bed" so you wouldn't have to get the entire name exactly right (case-sensitively, too).

theychx: Yes, but the problem is that, in the case of a cache miss, we are doing `get_chromecast(friendly_name=device_name)` in `CastController.init`. That would not be possible without the full name. Meaning that we then instead would need to do `get_chromecasts()` and then try and determine what cc the user ment. Not worth it if you ask me. Plus all the added complexity of needing to handle the possibility of multiple hits, we cant rely on the user making nice queries.

skorokithakis: Yes, we do need to always be looking up by `get_chromecasts()` and selecting the closest one, but we only have to select the first matching one (no complicated logic necessary). Either way, I'm not really hung up on this, so let's just leave it as-is for now.

catt/controllers.py
theychx: Fixed in new commit

skorokithakis:
Great work, thank you!

PR #11, opened 2016-12-11
Add config

theychx:
Howdy. As promised, here's the config.

There's a few things to consider with this PR:
I've had to change my strategy with regards to [this problem](https://github.com/pallets/click/issues/456#issuecomment-159543498), in order to make reading values from config into defaults work. Which also means that I've had to alter `entry_points` in `setup.py`.
Also, I realize that the code in `cli()` is suboptimal, as it is not the proper click-esque way of solving the problem, with raising that exception directly in the code. And I've been unsure as to whether I should pass the device value directly to `writeconfig()` from `cli()`, or to use the `pass_obj` decorator. I've chosen the latter, as for now, we can write ctx.obj to disk, as is.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -159,7 +159,7 @@ def writeconfig(settings):
     config_dir = click.get_app_dir("catt")
     try:
         os.mkdir(config_dir)
-    except FileExistsError:
+    except:
         pass
     config_filename = os.path.join(config_dir, "catt.json")
     with open(config_filename, "w") as config:
@@ -171,7 +171,7 @@ def readconfig():
     try:
         with open(config_filename, "r") as config:
             return json.load(config)
-    except FileNotFoundError:
+    except:
         return None
 
 
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -4,6 +4,7 @@ import random
 import os
 import socket
 import time
+import json
 from threading import Thread
 
 from .controllers import get_stream_info, CastController, Cache
@@ -16,29 +17,35 @@ def get_local_ip(cc_host):
     return s.getsockname()[0]
 
 
-class Device(object):
-    def __init__(self, name):
-        self.name = name
-
-
-pass_device = click.make_pass_decorator(Device)
+class CattCliError(click.ClickException):
+    pass
 
 
 @click.group()
-@click.option("--delete-cache", is_flag=True, help="Empty the Chromecast discovery cache.")
-@click.option("-d", "--device", metavar="NAME", help="Select Chromecast device.")
+@click.option("--delete-cache", is_flag=True,
+              help="Empty the Chromecast discovery cache.")
+@click.option("--write-config", is_flag=True,
+              help="Write name of default Chromecast device to config file")
+@click.option("-d", "--device", metavar="NAME",
+              help="Select Chromecast device.")
 @click.pass_context
-def cli(ctx, delete_cache, device):
+def cli(ctx, delete_cache, write_config, device):
     if delete_cache:
         Cache().clear()
-    ctx.obj = Device(device)
+    ctx.obj["device"] = device
+    if write_config:
+        if device:
+            cast = CastController(device)
+            writeconfig()
+        else:
+            raise CattCliError("No device specified.")
 
 
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url")
-@pass_device
-def cast(device, video_url):
-    cast = CastController(device.name)
+@click.pass_obj
+def cast(settings, video_url):
+    cast = CastController(settings["device"])
     cc_name = cast.cast.device.friendly_name
 
     if "://" not in video_url:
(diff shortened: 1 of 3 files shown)


theychx:
In `cli()` I am creating a `CastController` object, in order to ensure that the specified device actually exists on the network. That unfortunately makes flake complain. I'm not quite sure what to do about that.

skorokithakis:
Thank you for this! I haven't forgotten about it, I've just been very busy lately, but I'm going to get to it soon. I'm also trying to figure out how to solve the `flake8` problem, apart from the obvious, hacky `# noqa`.

catt/cli.py
```diff
@@ -16,29 +17,35 @@ def get_local_ip(cc_host):
...
+    pass
 
 
 @click.group()
-@click.option("--delete-cache", is_flag=True, help="Empty the Chromecast discovery cache.")
-@click.option("-d", "--device", metavar="NAME", help="Select Chromecast device.")
+@click.option("--delete-cache", is_flag=True,
+              help="Empty the Chromecast discovery cache.")
+@click.option("--write-config", is_flag=True,
+              help="Write name of default Chromecast device to config file")
```
skorokithakis: Missing full stop at the end, for consistency.

catt/cli.py
```diff
@@ -16,29 +17,35 @@ def get_local_ip(cc_host):
...
-def cli(ctx, delete_cache, device):
+def cli(ctx, delete_cache, write_config, device):
     if delete_cache:
         Cache().clear()
-    ctx.obj = Device(device)
+    ctx.obj["device"] = device
+    if write_config:
+        if device:
+            cast = CastController(device)
+            writeconfig()
```
skorokithakis: I think `writeconfig` should definitely accept the `settings` object here, rather than get it through `click`. I'm also not crazy about it being called `writeconfig` instead of `write_config`, but I can see the collision...

catt/cli.py
```diff
@@ -68,85 +75,109 @@ def cast(device, video_url):
...
+@click.pass_obj
+def writeconfig(settings):
+    config_dir = click.get_app_dir("catt")
+    try:
+        os.mkdir(config_dir)
+    except FileExistsError:
+        pass
+    config_filename = os.path.join(config_dir, "catt.json")
+    with open(config_filename, "w") as config:
+        json.dump(settings, config)
```
skorokithakis: I'm not loving JSON for configuration, I generally prefer ini-style files with the built-in `ConfigParser`.

skorokithakis:
Thanks again for the PR, I've left a few comments. Can you tell me a bit about the motivation behind this PR? I'm mainly concerned about the scattergun approach of saving all the config settings, I'm assuming you want to do this to avoid selecting a Chromecast every time?

If so, a better way might be to just add a config file that will be read with ConfigParser, like what I've [done here](https://gitlab.com/stavros/yeecli/blob/master/yeecli/cli.py), for example. The added advantage there would be that it would also allow for assignment of friendly names, so one could do `-d livingroom`, or something like that.

theychx:
Thanks for your feedback.

We can pass whatever we like to `writeconfig()`, so I'm not sure I understand your scattergun point. At the the moment `ctx.obj` only contains `settings`, thus my current approach. But I agree that it would be better not to pass the object to `writeconfig()` through the decorator.
My main motivation for using JSON was that it seemed like the simplest solution to the problem, and it also seemed to be easiest way to do --write-config, so that users do not have to fiddle with config-files themselves.
I realize that it is important to get this right in the first go, as changing the format of the config later will be a pain, so if you want this re-implemented using an ini config, I'll do that, no problem.

catt/cli.py
```diff
@@ -16,29 +17,35 @@ def get_local_ip(cc_host):
...
 @click.pass_context
-def cli(ctx, delete_cache, device):
+def cli(ctx, delete_cache, write_config, device):
     if delete_cache:
         Cache().clear()
-    ctx.obj = Device(device)
+    ctx.obj["device"] = device
+    if write_config:
+        if device:
+            cast = CastController(device)
```
theychx: As I spoke about in #12, we will need to implement `get_chromecast` ourselves. This would also solve this problem.

catt/cli.py
theychx: Agree. Let me know if you want another name for `writeconfig()`.

skorokithakis: If you can find a better one, it would be good, but I'm not too fussy about it staying the way it is.

skorokithakis:
The "scattergun" point was mainly because it will write whatever the current settings are (correct me if I'm mistaken), so if someone has specified `--delete-cache -w`, then the cache will always be deleted for them, right?

I agree that write-config is the easiest way to do it, but it's also kind of limiting in the long run. For example, you have multiple Chromecasts, so compare the two potential configs:

```
{"device": "Stavros's Chromecast", "aliases": {"livingroom": "Stavros's Chromecast", "bedroom": "Stavros's Other Chromecast"}}
```

```
device = Stavros's Chromecast

[aliases]
livingroom = Stavros's Chromecast
bedroom = Stavros's Other Chromecast
```

The former is rather harder to read/edit. If we only ever want to save some defaults with `write-config`, it's perfect, because the user doesn't have to edit it, but if we want to support other functionality later on (which won't be exposed through the CLI at all), JSON is going to make it harder.

This is the same as what I want to do for `yeecli`, where I want to have multiple bulbs (with nicknames) so people don't have to type the entire bulb name. There, I also want to have groups of bulbs, so people can say "turn on livingroom" and all N bulbs in the living room turn on. That's probably unnecessary here, but it's an example of configuration that's hard to express through the CLI, so one wants to make the config file easier to edit so the config can be expressed more easily.

What do you think? What's the use case that you, personally, find would benefit from a config file the most?

theychx:
You are mistaken about all options being written to disk, we only write what we put into `ctx.obj`, which at the moment only is `device`.
But I basically agree with all of your other points, so I guess it boils down to what the scope of `catt` is (your call). I have a few things lined up, but none of this would complicate this problem, and I'm not sure what would.
I think that the majority of users with more than one CC, would tend to use `catt` directly from the CLI with the same CC, so the config would mainly be there to make it easier to consistently adress the same device as a default, not to enable users to easily select a CC. 
Thats how i use `catt` myself at least. Whenever i use `catt` with my CC Audio, it is nearly always from my newsreader, where `catt` will play an enclosure when I double-click on it.

skorokithakis:
Ahh, okay, I've never used click contexts so I missed `ctx.obj`, thanks.

My use case for multiple devices is different from yours, as I have voice commands on my phone and say "play <whatever> on living room", so it looks up <whatever> in YouTube and casts that to the living room Chromecast (I don't actually have multiple Chromecasts now, but that's how my one works, and that's how my multiple bulbs work).

I do think an alias feature would be nice, as well as the default device. Luckily, the only change that needs to be made is using the `ConfigParser` library to read/write the file instead of `json` ([here](https://docs.python.org/2/library/configparser.html#examples)'s an example).

I can implement this myself, if you don't feel like doing double the work, just let me know. Otherwise, I think this is good to merge!

theychx:
OK , let's just go ahead with this now then, and do the alias stuff later.
But I might as well do `get_chromecast` now, to solve the flake problem, no? (before you merge)

skorokithakis:
Yeah, the alias stuff will come later, no problem. Yes, the `get_chromecast` change would be nice, to solve both the flake problem and the filtering being taken out. Are you going to do the ConfigParser change too, or should I?

theychx:
I would prefer not to do the `ConfigParser` stuff now, in order to get this out of the way

catt/cli.py
theychx: fixed

skorokithakis:
Sure thing, just let me know when you're done.

theychx:
So this should be done. There is a sligh catch, in that we now are doing superfluous auto-discovery when the user is doing `--write-config`. But that is a minor problem after all, and I have no idea on how to solve it, without completely reworking `controllers.py`, so I hope it is ok.

skorokithakis:
That's an acceptable tradeoff, thank you for your help!

PR #13, opened 2016-12-19
feat: Add custom time type for seek functions

theychx:
Hi. I have made this little handy custom parameter type (as described [here](http://click.pocoo.org/6/parameters/#implementing-custom-types)), that enables the user to use normal time descriptions when using `seek`, `ffwd` or `rewind`. 

F.ex. `catt seek 01:14:20`.
Something like `catt ffwd 500` would still work as is does now.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -21,6 +21,22 @@ class CattCliError(click.ClickException):
     pass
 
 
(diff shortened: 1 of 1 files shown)


skorokithakis:
Looks fantastic, thank you!

PR #14, opened 2016-12-19
fix: Make catt exit nicely if ytdl does not find a video.

theychx:
Another small fix

In case you are planning on doing a release one of these days, I have two more minor PR's coming up:

+ Slight cleanup in `controllers.py`
+ Remove volume default in `cli.py`


skorokithakis:
Thanks, I hit that error the other day as well, but forgot to fix it.

PR #15, opened 2016-12-20
fix: Remove volume default.

theychx:
`volume` should expect an argument, so I removed the default.


skorokithakis:
While we're at it, maybe we should multiply and clamp it to 0-100, rather than 0-1?

theychx:
Good idea. I'll do it right away.

skorokithakis:
Thank you!

theychx:
Done

PR #17, opened 2016-12-20
fix: Cleanup in controllers / cli.

theychx:
Looks nicer and works as expected.

.......and that should be the end of this series of PR's.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -118,7 +118,6 @@ class CastController:
             cache.set(self.cast.name, self.cast.host)
         time.sleep(0.2)
 
-
     def play_media(self, url, content_type="video/mp4"):
         self.cast.play_media(url, content_type)
 
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -105,8 +105,6 @@ class Cache:
 
 class CastController:
     def __init__(self, device_name):
-        if device_name:
-            echo("Trying to connect to %s" % device_name)
         cache = Cache()
         cached_ip = cache.get(device_name)
 
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -26,6 +26,20 @@ def get_stream_info(video_url):
     return stream_info
 
 
+def get_chromecast(device_name):
+    devices = pychromecast.get_chromecasts()
+    if device_name:
+        try:
+            return next(cc for cc in devices if cc.name == device_name)
+        except StopIteration:
+            raise CattCastError("Specified device not found.")
+    else:
+        try:
+            return min(devices, key=lambda cc: cc.name)
+        except ValueError:
+            raise CattCastError("No devices found.")
+
+
 class CattCastError(ClickException):
     pass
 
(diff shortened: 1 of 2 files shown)


catt/controllers.py
```diff
@@ -102,25 +114,10 @@ def __init__(self, device_name):
...
         except (pychromecast.error.ChromecastConnectionError, ValueError):
-            if device_name:
-                self.cast = self.get_chromecast(friendly_name=device_name)
-            else:
-                self.cast = self.get_chromecast()
+            self.cast = get_chromecast(device_name)
             cache.set(self.cast.name, self.cast.host)
         time.sleep(0.2)
 
-    def get_chromecast(self, friendly_name=None):
```
skorokithakis: Ah, I meant to investigate: How did this even work, if it wasn't called on an instance? I can only guess `self` was actually the class argument there.

catt/controllers.py
```diff
@@ -26,6 +26,20 @@ def get_stream_info(video_url):
     return stream_info
 
 
+def get_chromecast(device_name):
+    devices = pychromecast.get_chromecasts()
+    if device_name:
```
skorokithakis: Why not something like this?:

~~~.python
# Sort consistently.
devices = sorted(pychromecast.get_chromecasts(), key=lambda cc: cc.name)
if device_name:
    devices = [device for device in devices if device.name == device_name]
if not devices:
    raise CattCastError("No suitable device found.")
return devices[0]
~~~

theychx: That looks nice, I just think that for testing purposes, it is useful to be able to distinguish between "specified not found" and "none found". But we can do it like you suggest if you prefer.

catt/controllers.py
theychx: That was a screwup on my part, sorry.

catt/controllers.py
skorokithakis: It doesn't distinguish though, does it? If you specify a device name it always says "Specified not found", otherwise "None found". If you specify a device name and it finds no devices, it will still say "Specified not found". Or do you mean in case you specified a device in the config and forgot about it? It could be useful then, I agree.

catt/controllers.py
skorokithakis: It's fine, I was just wondering.

catt/controllers.py
skorokithakis: Sure thing, but I think we should at least print the device name when it has been specified, so people know that it's in the configuration. So, if you've specified a device, we should print at the start "Trying to connect to <name>..." or something like that.

theychx: Looking at your code again, I see that you are sorting `devices`, regardless of whether it's needed or not, which I do not like. So I ended up reworking my own code, to make it work properly.

skorokithakis: Sorting devices is always needed, because you don't want to rely on the network-returned order, even when you're specifying a device name, because we might want to do an `in` later on. This is fine for now, though, I can change it when/if I implement the `in`.

PR #18, opened 2016-12-22
fix: Replace connection message with more useful error message.

theychx:
Less annoying and just as useful.


skorokithakis:
Thanks!

PR #19, opened 2016-12-22
feat: Add aliases to config.

theychx:
This PR is a bit of a POC at this point, but it seems to be working rather well.
Note that I have used the name `"defaults"` for the default section, as the values from `"DEFAULT"` are read into all sections, which is not useful in this scenario.
`main()` does not look too pretty, so if you know how to make it more concise then that would be great.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -14,6 +14,10 @@ from .controllers import get_stream_info, get_chromecast, CastController, Cache
 from .http_server import serve_file
 
 
+CONFIG_DIR = click.get_app_dir("catt")
+CONFIG_FILENAME = os.path.join(CONFIG_DIR, "catt.cfg")
+
+
 def get_local_ip(cc_host):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     s.connect((cc_host, 0))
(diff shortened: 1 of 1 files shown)


skorokithakis:
I mean to review this, sorry about the delay (holidays are hectic)!

theychx:
No worries.

skorokithakis:
Hm. I made some changes on this and wanted to push them to your branch so you still get credit, but I don't have write access. I pushed the changes into an add-aliases branch here, should I open a PR for that and continue the discussion with you there, or do you want to give me write access to your `catt` repo so I can push the changes there and continue in this PR?

theychx:
It's fine with the new branch, I don't care about the credit.
But I don't understand about you not having write-access, I thought thats what "Allow edits from maintainers." were for.

skorokithakis:
Ah, you're right, it lets me make changes on the web UI, but I can't push my changes to your branch from my local git. I'm probably doing something wrong.

skorokithakis:
Alright, turns out the HTTPS transport doesn't like me. What do you think of the changes?

theychx:
`readconfig` looks great, kind of what I wanted to do all along. But there is something wrong with `writeconfig`, `catt -d name write_config` fails with a clean slate.

skorokithakis:
Good catch, (hopefully) fixed, thanks.

skorokithakis:
If it's okay with you, I can merge this.

theychx:
It seems to be working great now, so thats ok.

Issue #23, opened 2017-03-13
Youtube playlists 


blomdahldaniel:
When I provide catt with a youtube link to a youtube playlist, the youtube dl is triggered correct (as it seems) but catt is faling. This is the logg for when i try to do this.

So, is this a bugg just form me or is thre functionality just not there yet for playlists?

I am on macOS and all other features in `catt` is working properly.

Logg
```bash
▶ catt cast https://www.youtube.com/watch\?v\=kk0WRHV_vt8\&list\=PL52RKVKBFM328OwddSMRQS9BYOA3oVDgl

Casting remote file https://www.youtube.com/watch?v=kk0WRHV_vt8&list=PL52RKVKBFM328OwddSMRQS9BYOA3oVDgl...
[youtube:playlist] Downloading playlist PL52RKVKBFM328OwddSMRQS9BYOA3oVDgl - add --no-playlist to just download video kk0WRHV_vt8
[youtube:playlist] PL52RKVKBFM328OwddSMRQS9BYOA3oVDgl: Downloading webpage
[download] Downloading playlist: Snarky Puppy - We Like It Here
[youtube:playlist] playlist Snarky Puppy - We Like It Here: Downloading 8 videos
[download] Downloading video 1 of 8
[youtube] kk0WRHV_vt8: Downloading webpage
[youtube] kk0WRHV_vt8: Downloading video info webpage
[youtube] kk0WRHV_vt8: Extracting video information
[youtube] kk0WRHV_vt8: Downloading MPD manifest
[download] Downloading video 2 of 8
[youtube] fuhHU_BZXSk: Downloading webpage
[youtube] fuhHU_BZXSk: Downloading video info webpage
[youtube] fuhHU_BZXSk: Extracting video information
[youtube] fuhHU_BZXSk: Downloading MPD manifest
[download] Downloading video 3 of 8
[youtube] kDXnPfA_5pY: Downloading webpage
[youtube] kDXnPfA_5pY: Downloading video info webpage
[youtube] kDXnPfA_5pY: Extracting video information
[youtube] kDXnPfA_5pY: Downloading MPD manifest
[download] Downloading video 4 of 8
[youtube] eFUOZ5PI3m8: Downloading webpage
[youtube] eFUOZ5PI3m8: Downloading video info webpage
[youtube] eFUOZ5PI3m8: Extracting video information
[youtube] eFUOZ5PI3m8: Downloading MPD manifest
[download] Downloading video 5 of 8
[...]

PR #24, opened 2017-03-14
fix: Better handling of playlists

theychx:
As discussed in #23.

Tested with youtube, bandcamp, nfl and generic, but probably needs a bit more testing.

EDIT: ready to merge


skorokithakis:
This is ideal, thank you, @theychx. You should add your name to the `AUTHORS` file, too.

Issue #25, opened 2017-03-15
Feature request: List of Chromecasts available


blomdahldaniel:
A command that echoes out the names (and statuses, playing/not playing) of the available chromecasts. 

Would be a great feature!

theychx:
I'm most in favour of doing this, with just the information we get from auto-discovery - which is: name, device type, ip-adress. Otherwise a good idea.

skorokithakis:
Agreed, it'll help with the multiple Chromecast feature. I'll do that today.

skorokithakis:
Damnit, I meant to make a PR for @theychx to review but botched the branch. @theychx, can you tell me if e9e3fb9089ad3335a2190f47845d7f7aedcdc4eb looks okay?

theychx:
Works great.
But wouldn't it be easier to make a `CastController.scan` method that calls `pychromecast.get_chromecasts` and echo's, and just call that from `cli.py`, like the rest of the functionality?

skorokithakis:
Fixed in e9e3fb9089ad3335a2190f47845d7f7aedcdc4eb.

skorokithakis:
@theychx Oh, have they committed to that? I don't see anything in the documentation, and I'd rather not rely on an implementation detail. Can you point me to the docs where they define `get_chromecasts` as sorted?

theychx:
Fair point. And sorry about bringing this up in the other PR-thread, that was a bit obnoxious. :pig:

skorokithakis:
Nah, no problem, it could have saved us a few lines of code.

PR #28, opened 2017-03-18
fix: Make catt work with pychromecast 0.8.0

### LINKED ISSUE #12: catt will break with next pychromecast release
Upstream has removed `get_chromecast` from pychromecast, see https://github.com/balloob/pychromecast/commit/dac71620f6faa22a2da79a2432eb9126f30292e1.
So we need to either freeze the pychromecast requirement or implement auto-discovery filtering ourselves. I can do a PR for either, if you are interrested.

theychx:
Howdy.
So I worked around the issues and made this work almost perfectly (see notes).

With this PR, `catt` will now fail if nothing is playing (except `cast` of course), whereas before, you could do `catt ffwd 50` when nothing was playing, and `catt` would just exit quietly. The failling behaviour could also be useful in scripts.

Fixes #12 
EDIT: ready to merge

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -133,7 +133,6 @@ class Cache:
 
 class CastController:
     def __init__(self, device_name, state_check=True):
-        INACTIVE = ["UNKNOWN", "IDLE"]
         cache = Cache()
         cached_ip = cache.get(device_name)
 
@@ -148,11 +147,14 @@ class CastController:
         self.cast.wait()
 
         if state_check:
-            if self.cast.app_id == "E8C28D3C" or not self.cast.app_id:
-                raise CattCastError("Chromecast is inactive.")
-            self.cast.media_controller.block_until_active(1.0)
-            if self.cast.media_controller.status.player_state in INACTIVE:
-                raise CattCastError("Nothing is currently playing.")
+            self._check_state()
+
+    def _check_state(self):
+        if self.cast.app_id == "E8C28D3C" or not self.cast.app_id:
+            raise CattCastError("Chromecast is inactive.")
+        self.cast.media_controller.block_until_active(1.0)
+        if self.cast.media_controller.status.player_state in ["UNKNOWN", "IDLE"]:
+            raise CattCastError("Nothing is currently playing.")
 
     def play_media(self, url, content_type="video/mp4"):
         self.cast.play_media(url, content_type)
@@ -163,9 +165,6 @@ class CastController:
 
     def pause(self):
         self.cast.media_controller.pause()
-
-    def stop(self):
-        self.cast.media_controller.stop()
 
     def seek(self, seconds):
         self.cast.media_controller.seek(seconds)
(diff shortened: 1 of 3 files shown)


catt/controllers.py
```diff
@@ -143,30 +143,45 @@ def __init__(self, device_name):
...
+        self.cast.media_controller.block_until_active(1.0)
+        if self.cast.media_controller.status.player_state in ["UNKNOWN", "IDLE"]:
+            raise CattCastError("Nothing is currently playing.")
 
     def play_media(self, url, content_type="video/mp4"):
         self.cast.play_media(url, content_type)
+        self.cast.media_controller.block_until_active()
 
     def play(self):
+        self._check_state()
```
skorokithakis: I'm not in love with calling `_check_state()` in every command. Maybe run it in the constructor and add a `no_state_check` parameter for when we don't want it? That way we only need to specify that parameter once, and the state checks are removed from all the other calls, and we can still use `self.seek`.

theychx: Good point. Let me work on that.

catt/controllers.py
```diff
@@ -146,42 +147,36 @@ def __init__(self, device_name):
 
         self.cast.wait()
 
-    def _check_state(self):
-        if self.cast.app_id == "E8C28D3C" or not self.cast.app_id:
-            raise CattCastError("Chromecast is inactive.")
-        self.cast.media_controller.block_until_active(1.0)
-        if self.cast.media_controller.status.player_state in ["UNKNOWN", "IDLE"]:
-            raise CattCastError("Nothing is currently playing.")
+        if state_check:
```
skorokithakis: I quite liked this as a separate function (it's better documentation), but this is fine as well.

theychx: I can change it back if you prefer, just let me know.
Another thing: In cli we have a `stop` command which is calling `CastController.kill`, and then there is a `CastController.stop` which is currently unused. Do you want to adress that?

skorokithakis: It's generally better to have code that does one thing as a separate function, so they're more self-contained and better documented. If you don't mind changing it, that'd be great, but if you can't be bothered, this is fine as well.

I used the `kill` command instead of the `stop` command because the latter didn't go back to the main Chromecast screen, and I think that would be more intuitive for the user. What's your opinion?

theychx: I change it back right away.
I agree about stop/kill, but it's that there is an unused method in `CastController`. I think we should remove it, as we are unlikely to use it.

theychx: Done. I removed `CastController.stop`.

skorokithakis:
Looks good to me, thanks!

theychx:
By the way: Did you forget about #25?

skorokithakis:
I forgot to link it in e9e3fb9089ad3335a2190f47845d7f7aedcdc4eb, it's implemented as `catt scan`.

theychx:
> I forgot to link it in e9e3fb9, it's implemented as catt scan

Yes, I noticed that, I was referring to https://github.com/skorokithakis/catt/issues/25#issuecomment-286990642.

PR #29, opened 2017-03-19
feat: Add argument to volumeupdown

theychx:
Now that we are using `pychromecast 0.8.0`, we can make use of a new feature: The ability to supply a delta to `volume_up` and `volume_down`. I have also used the opportunity to make the type of the value consistent with what we are using in `volume`.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -172,7 +172,7 @@ def volume(settings, level):
     cast.volume(level / 100.0)
 
 
-@cli.command(short_help="Turn up volume by an DELTA increment.")
+@cli.command(short_help="Turn up volume by a DELTA increment.")
 @click.argument("delta", type=click.IntRange(1, 100),
                 required=False, default=10, metavar="DELTA")
 @click.pass_obj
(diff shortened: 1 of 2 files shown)


catt/cli.py
```diff
@@ -172,18 +172,22 @@ def volume(settings, level):
     cast.volume(level / 100.0)
 
 
-@cli.command(short_help="Turn up volume by an 0.1 increment.")
+@cli.command(short_help="Turn up volume by an DELTA increment.")
```
skorokithakis: Typo: "a" instead of "an".

catt/cli.py
```diff
@@ -172,18 +172,22 @@ def volume(settings, level):
...
+@click.argument("delta", type=click.IntRange(1, 100),
+                required=False, default=10, metavar="DELTA")
 @click.pass_obj
-def volumeup(settings):
+def volumeup(settings, delta):
     cast = CastController(settings["device"], state_check=False)
-    cast.volumeup()
+    cast.volumeup(delta / 100.0)
```
skorokithakis: Why the float? Shouldn't it be an int?

catt/cli.py
theychx: fixed

catt/cli.py
theychx: Nope.
```
Signature: cast.volume_up(delta=0.1)
Docstring:
Increment volume by 0.1 (or delta) unless it is already maxed.
Returns the new volume.
File:      ~/python/chromecast/lib/python3.5/site-packages/pychromecast/__init__.py
Type:      method
```

skorokithakis: Oh, right, oops, thanks.

PR #31, opened 2017-03-23
fix: Further improve play_media

theychx:
Hi. So while experimenting with listeners, I found out that `catt cast` does not work if run when the cc is already playing something (in an app other than the "Default Media Receiver"). I discussed this briefly with the friendly pychromecast people: https://github.com/balloob/pychromecast/issues/168

So I did my own listener that sets its own event, to use instead of `block_until_active` in `CastController.play_media`. I realize that this is getting a bit involved, but as you can see in the before mentioned discussion, `block_until_active` simply does not cut it in it's current form, for use in `CastController.play_media`.

EDIT: Sorry about the extra commits, my own flake8 did not complain about import order. And now CI is failing on python3, I have no idea why (the branch installs fine).

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import tempfile
 import time
+import threading
 
 import pychromecast
 import youtube_dl
@@ -131,8 +132,18 @@ class Cache:
             pass
 
 
+class MediaStatusListener:
+    def __init__(self):
+        self.ready = threading.Event()
+
+    def new_media_status(self, status):
+        if status.player_state == "BUFFERING":
+            self.ready.set()
+
+
 class CastController:
     def __init__(self, device_name, state_check=True):
+        self.listener = MediaStatusListener()
         cache = Cache()
         cached_ip = cache.get(device_name)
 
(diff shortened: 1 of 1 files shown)


skorokithakis:
No problem about the commits, we're squashing anyway. The test failure seems like this old netifaces issue: https://bitbucket.org/al45tair/netifaces/issues/13/0104-install-is-broken-on-python-3x

I thought that was resolved, though, hmm...

skorokithakis:
Yeah, Pychromecast needs to use netifaces 0.10.5, earlier versions break on Python 3.

theychx:
I've changed this once again, to make it even more bulletproof, and to make it more consistent with the advice given by armills.

skorokithakis:
Looks good to me, although I'm very curious about the failing tests... I wonder why Pychromecast doesn't use the new netifaces version, I should ask them.

theychx:
The python-zeroconf guy changes the netifaces requirement to <=0.10.4 because of some Windows problem.

skorokithakis:
Ah :/ That's too bad, that version has a Python 3 problem...

theychx:
The thing is, it installs fine on python 3. What is the issue exactly?

skorokithakis:
The issue is described [here](https://bitbucket.org/al45tair/netifaces/issues/13/0104-install-is-broken-on-python-3x) and fixed [here](https://bitbucket.org/al45tair/netifaces/pull-requests/7/fix-setup-in-case-theres-no-print/diff), basically netifaces won't install on platforms where `print` isn't available (which happens much more frequently than you'd think).

theychx:
Hmmm, thank you for the info.

PR #32, opened 2017-03-26
feat: Make status use regular time descriptions

theychx:
Hi. Just a little adjustment, so that the output of `status` is consistent with the time descriptions the user can use with `seek` etc.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -207,25 +207,22 @@ class CastController:
     def status(self):
         status = self.cast.media_controller.status.__dict__
 
-        if not status["duration"]:
-            echo("State: {player_state}\n".format(**status))
-            return
+        if status["duration"]:
+            dur, cur = int(status["duration"]), int(status["current_time"])
+            duration = time.strftime('%H:%M:%S', time.gmtime(dur))
+            current = time.strftime('%H:%M:%S', time.gmtime(cur))
+            remaining = time.strftime('%H:%M:%S', time.gmtime(dur - cur))
+            progress = int((1.0 * cur / dur) * 100)
 
-        status["current_time"] = int(status["current_time"])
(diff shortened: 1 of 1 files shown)


catt/controllers.py
```diff
@@ -207,25 +207,22 @@ def volumedown(self, delta):
     def status(self):
         status = self.cast.media_controller.status.__dict__
 
-        if not status["duration"]:
-            echo("State: {player_state}\n".format(**status))
-            return
+        if status["duration"]:
+            dur, cur = int(status["duration"]), int(status["current_time"])
+            duration = time.strftime('%H:%M:%S', time.gmtime(dur))
```
skorokithakis: You may want:

~~~
human_time = lambda x: time.strftime('%H:%M:%S', time.gmtime(x))
~~~

Or use a normal def, as it's clearer.

theychx: fixed

skorokithakis:
Thank you!

theychx:
Please consider making a release one of these days, so that users can upgrade PyChromecast.

skorokithakis:
I'm kind of waiting for a version that won't break in Python 3 :/ Do you think they'll be upgrading netifaces?

theychx:
It's really out of the hands of the PyChromecast people, as this is caused by a decision made by the person behind python-zeroconf (PyChromcast is not directly depending on netifaces). So the issue persists for new users, regardless of what we do.

skorokithakis:
Bah, yeah, we depend on pychromecast, they depend on zeroconf, they depend on netifaces, and the latter breaks... Alright, I'll make a release one of these days, thank you.

theychx:
Yeah this sucks, as the development pace of netifaces seems to be very slow, so this might persist for a long time. Let's cross our fingers.

PR #33, opened 2017-06-27
fix: Turn human_time into a def

theychx:
Howdy.
So the zeroconf/netifaces issues were resolved, giving us the opportunity to make CI builds pass once again. Running flake on `controllers.py` resulted in:
```
controllers.py:211:13: E731 do not assign a lambda expression, use a def
```
so I obeyed its command.


skorokithakis:
Hoisted by my own petard!

PR #34, opened 2017-06-27
Another slight improvement to play_media

theychx:
This is probably not going to make any difference in real use, but IMO it is better to register the listener before calling `media_controller.play_media`. With the current implementation there is a (very) slight chance that `self.listener.ready.wait()` will lock indefinitely.


skorokithakis:
Looks good to me, thank you!

PR #36, opened 2017-06-28
Handle extractor fail

theychx:
Here's a link that makes `catt` crash: http://www.gdcvault.com/play/1014846/Conference-Keynote-Shigeru

I'm not entirely shure we need to check for `ExtractorError`, but best to leave it there.


skorokithakis:
Looks good, thanks!

PR #39, opened 2017-07-01
Summer cleaning

theychx:
Hi. Just a bit of tidying up here. A few variable names that were too short or redefined something builtin or from outer scope. Fixed that. Also added quotes to device name in error message, that could otherwise be confusing.
I've tested all functionality and everything seems to work.


skorokithakis:
Thank you!

PR #40, opened 2017-07-05
[WIP] YouTube Queue support

theychx:
~~Use [this fork of pychromecast](https://github.com/theychx/pychromecast) with this branch.~~

In case you're wondering, the controller does unfortunately not support dumping a yt playlist in one go (this works when casting from my Android phone), thus the lame method of adding videos one by one.

### DIFF
--- a/catt/youtube.py
+++ b/catt/youtube.py
@@ -4,7 +4,10 @@ Controller to interface with the YouTube-app.
 
 import re
 import threading
-from json import JSONDecodeError
+try:
+    from json import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
 
 import requests
 
--- a/catt/youtube.py
+++ b/catt/youtube.py
@@ -24,7 +24,7 @@ BASE_REQUEST_PARAMS = {"device": "REMOTE_CONTROL", "id": RANDOM_ID, "name": "Des
                        "CVER": 1}
 
 SET_PLAYLIST_METHOD = {"method": "setPlaylist", "params": VIDEO_ID_PARAM, "TYPE": None}
-REQUEST_PARAMS_SET_PLAYLIST = {**BASE_REQUEST_PARAMS, **SET_PLAYLIST_METHOD}
+REQUEST_PARAMS_SET_PLAYLIST = dict(**dict(BASE_REQUEST_PARAMS, **SET_PLAYLIST_METHOD))
 
 REQUEST_DATA_SET_PLAYLIST = "count=0"
 REQUEST_DATA_ADD_TO_PLAYLIST = "count=1&ofs=%d&req0__sc=addVideo&req0_videoId=%s"
--- a/catt/youtube.py
+++ b/catt/youtube.py
@@ -1,0 +1,381 @@
+"""
+Controller to interface with the YouTube-app.
+"""
+
+import re
+import threading
+from json import JSONDecodeError
+
+import requests
+
+from pychromecast.controllers import BaseController
+
+YOUTUBE_BASE_URL = "https://www.youtube.com/"
+YOUTUBE_WATCH_VIDEO_URL = YOUTUBE_BASE_URL + "watch?v="
+
+# id param is const(YouTube sets it as random xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx so it should be fine).
+RANDOM_ID = "12345678-9ABC-4DEF-0123-0123456789AB"
+VIDEO_ID_PARAM = '%7B%22videoId%22%3A%22{video_id}%22%2C%22currentTime%22%3A5%2C%22currentIndex%22%3A0%7D'
+TERMINATE_PARAM = "terminate"
+
+REQUEST_URL_SET_PLAYLIST = YOUTUBE_BASE_URL + "api/lounge/bc/bind?"
+BASE_REQUEST_PARAMS = {"device": "REMOTE_CONTROL", "id": RANDOM_ID, "name": "Desktop&app=youtube-desktop",
+                       "mdx-version": 3, "loungeIdToken": None, "VER": 8, "v": 2, "t": 1, "ui": 1, "RID": 75956,
+                       "CVER": 1}
+
+SET_PLAYLIST_METHOD = {"method": "setPlaylist", "params": VIDEO_ID_PARAM, "TYPE": None}
+REQUEST_PARAMS_SET_PLAYLIST = {**BASE_REQUEST_PARAMS, **SET_PLAYLIST_METHOD}
+
+REQUEST_DATA_SET_PLAYLIST = "count=0"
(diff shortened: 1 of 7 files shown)


skorokithakis:
Is this something we should merge? It only works with the fork right now, right?

theychx:
> It only works with the fork right now, right?

Yep, I've tried to get in contact with ur1katz (author of the new controller), through mentioning him in [this](https://github.com/balloob/pychromecast/issues/62) pychromecast issue. So far he has not responded.
> Is this something we should merge?

Looking at the code now, I think there is too much assuming going on in ``cli.cast``. The overall approach is good, but I'm having second thoughts about ``StreamInfo`` magically returning the most useful data. I would greatly value your opinion on this, as the code (as it is now) seems to be working great.

skorokithakis:
Thanks, I'll give it a look regardless and leave comments in a review. Thanks for all your work!

skorokithakis:
I read through the changes and they all seem very sane and straightforward. I'm very excited about playlist support, and it seems like you've implemented it in a nice way. All this PR needs to be merged is the upstream merge of ur1katz's branch and any changes for any refactors they might do, correct?

theychx:
Yes exactly. Please note that the `youtube.py` controller itself is the only difference between ur1katz' branch and baloobs. So if upstream doesn't want to merge (as the controller is prone to break again), we could import it directly into `catt`. We would still need ur1katz' feedback though (or?).

skorokithakis:
Could we just vendor one of the models like that? Wouldn't the rest of pychromecast that called it break? Or is it not called from anywhere?

If it's called from the rest of pychromecast, we'd have to monkey-patch it in. I'd be more comfortable with ur1katz' feedback, but we can certainly merge the file without it if we see that it works well enough...

skorokithakis:
About the `StreamInfo` magic, what I usually do is make a function and call it `return_most_useful_data` or something similar, so it's clear what's going on. That serves as documentation and decreases the amount of magic.

theychx:
> Could we just vendor one of the models like that?

Yep. Please see [this](https://github.com/ur1katz/pychromecast/blob/master/README.rst).
Notice `register_handler`. You can even subclass the build-in MediaController, and then register an object created from it, to interace with the `Stylized Media Receiver` (tried it yesterday with BubbleUPnP SMR, works great).

In the event that ur1katz remains MIA, can we just use his controller or how does that work? Doesn't his fork inherit the license of main project?

skorokithakis:
Yes, he left the MIT license in his repo so that's the license. We're using the BSD license, and the two are compatible, AFAIK, so we can use that file as long as we credit the author in it.

theychx:
Do you want to go ahead with that plan now then? He hasn't been active here on GitHub since february, so we could otherwise be in for a long wait.

skorokithakis:
Oh, yeah, if he has been inactive for this long, sure. We should test it a bit after merging (and I'll test it on my setup as well), and if everything is good we can make a new release.

theychx:
I need to fix the tests. Forgot about that.

catt/youtube.py
```diff
@@ -0,0 +1,381 @@
...
+VIDEO_ID_PARAM = '%7B%22videoId%22%3A%22{video_id}%22%2C%22currentTime%22%3A5%2C%22currentIndex%22%3A0%7D'
+TERMINATE_PARAM = "terminate"
+
+REQUEST_URL_SET_PLAYLIST = YOUTUBE_BASE_URL + "api/lounge/bc/bind?"
+BASE_REQUEST_PARAMS = {"device": "REMOTE_CONTROL", "id": RANDOM_ID, "name": "Desktop&app=youtube-desktop",
+                       "mdx-version": 3, "loungeIdToken": None, "VER": 8, "v": 2, "t": 1, "ui": 1, "RID": 75956,
+                       "CVER": 1}
+
+SET_PLAYLIST_METHOD = {"method": "setPlaylist", "params": VIDEO_ID_PARAM, "TYPE": None}
+REQUEST_PARAMS_SET_PLAYLIST = {**BASE_REQUEST_PARAMS, **SET_PLAYLIST_METHOD}
```
theychx: This syntax was introduced in python3.5, making ci-builds fail on older pythons. I'm unsure how to fix this.

theychx: fixed

theychx:
So it would appear that
```
from json import JSONDecodeError
```
in `youtube.py` only works in python3.5. For older pythons you need to do
```
from simplejson import JSONDecodeError
```
That is an extra dependency, and what's worse, I'm not sure they are exchangeable, as some python2.7 youtube tests revealed some odd behaviour.

skorokithakis:
What error does 2.7's JSON library raise when there's a decoding problem? That's what we need to catch.

theychx:
That's the thing, it doesn't raise anything. It just behaves less reliably when casting youtube playlists, than what I'm used to with python3.5.
I did however at one point get this very strange error:
```
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner       
  File "/home/benno/.local/lib/python2.7/site-packages/pychromecast/socket_client.py", line 339, in run
  File "/home/benno/.local/lib/python2.7/site-packages/pychromecast/socket_client.py", line 375, in run_once
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'SSLError'
```
but that seems unrelated.

theychx:
Did you mean that we should test the code without the import statement?

skorokithakis:
No, I'm just confused. Why do we need to import something that doesn't get raised? I don't understand the problem, I think.

theychx:
https://github.com/skorokithakis/catt/blob/ytqueue-poc/catt/youtube.py#L168

That's the only place it's used. So far it has never thown the `YoutubeSessionError` at me, so I'm just as confused.
I see now that a new version of pychromecast was just released, so the odd behaviour could come from that.

skorokithakis:
Probably, at least I don't see how it could be otherwise... The missing exception doesn't seem very relevant to the problem.

theychx:
Should we remove the import statement and the try except clauses, and go on from there?

skorokithakis:
Python 2.x raises `ValueError`, so if importing fails you can just do `JSONDecodeError = ValueError` to have that work for 2.x.

theychx:
Should be ready to merge now. Sorry about the confusion, I think there was too many things that went wrong at the same time, here on my end.

skorokithakis:
Hey, no problem, take your time. I will (try to) approve it, merge when you feel comfortable.

skorokithakis:
Hey @theychx, can we add a line of credit to the authors file to @ur1katz, whose code we used here?

skorokithakis:
Oh good, he emailed me about it and I only looked in the authors, it didn't occur to me to look in the license. Thank you.

theychx:
> he emailed me
Really? Did he say anything else? I'd be really enthutiastic about him continuing his development of the controller.

skorokithakis:
No, he just said he was away and that he's happy we found some use for the controller he didn't get to finish. Hopefully he'll be around now!

PR #42, opened 2017-07-27
Add preliminary cc type detection

### LINKED ISSUE #41: Some Vimeo videos do not work anymore
So while fiddling with tests, I discovered that the url that the format selector returns for some vimeo videos, does not play on my cc 2. Turns out that the resolution of the `"best"` format is too high.
Replacing `"best"` with `"best[width <=? 1920][height <=? 1080]"` appears to solve the problem.
The cc ultra might be able to play the higher resolution format, but I do not have access to one of those.

theychx:
fixes #41

Now we just need to know what to look for in `cast.model_name` (or whatever). Until then, this PR will effectively force 1080p on everything.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -83,7 +83,7 @@ def cast(settings, video_url):
     cst = CastController(settings["device"], state_check=False)
     cc_name = cst.cast.device.friendly_name
     cc_type = cst.cast.cast_type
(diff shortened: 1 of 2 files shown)


skorokithakis:
Looks good to me, I haven't tried the Mi Box yet, but I'll tell you when I get home (should be Monday).

PR #43, opened 2017-07-27
Add skip command

theychx:
Since the youtube controller currently does not support selecting which playlist entry to play, I thought that a simple `skip` command would be useful. When casting a youtube playlist this would make the app play the next entry (if any). The implementation is a bit of a hack, but it seems to work well.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -204,6 +204,13 @@ def seek(settings, timedesc):
     cst.seek(timedesc)
 
 
+@cli.command(short_help="Skip to end of video.")
+@click.pass_obj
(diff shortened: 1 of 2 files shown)


skorokithakis:
Looks good to me, which part do you think is a hack? The one that checks for YouTube?

theychx:
> Looks good to me, which part do you think is a hack?
Seeking to the end of the video + 1 second.

skorokithakis:
Ah, that's not too terrible, for lack of a better solution. Maybe we should add a comment so we can replace it later, but it's good for now.

PR #47, opened 2017-08-23
Ensure DIAL info in get_chromecasts

theychx:
Hi. I think I took the wrong approach in my last PR, with regards to ensuring DIAL info.
In this PR, we take action in `get_chromecasts` instead of the `CastController` constructor. This also has the added benefit of making `catt scan` return DIAL info, which will be useful if users report back with information about 4K-capable devices.


skorokithakis:
Seems saner, thank you.

PR #48, opened 2017-08-23
Remove readthedocs entry from readme

theychx:
Seems a bit superflous anyway (if it existed :smiley: ).


skorokithakis:
Oops, good catch!

PR #49, opened 2017-08-23
Make audio devices stream DASH-audio from YouTube

theychx:
Hey. With YouTube offering audio-only DASH streams, we can utilize `StreamInfo` to enable this for audio devices, saving users a bit of bandwith.


skorokithakis:
It's more than "a bit", it's quite significant. Thanks for the PR!

PR #53, opened 2017-09-17
Further improve format selection

theychx:
Hi. Youtube-dl already does fallback format selection, so no need for us to implement that.


skorokithakis:
Thank you!

PR #54, opened 2017-09-30
Improve standard format

theychx:
Hey. A little 60fps tweak for twitch live streams (and content where framerate is reported by the extractor) on regular cc's.

It would be nice if you could make a release so that users can benefit from the YouTube functionality.


skorokithakis:
Released 0.5.2 (tooling failed and skipped 0.5.1, unfortunately), thank you!

PR #55, opened 2017-10-25
[WIP] Controllers refactor

theychx:
Hi
So while I did the YouTube stuff, I realized that ```catt``` would turn into a mess sooner rather than later, if we were to just continue to latch new functionality from different controllers onto existing code. So I propose this design, that tries to make the capabilities (or lack thereof) of any controller, transparent to the cli. Let me know what you think.

List of things that are not implemented yet:
+ ~~Playing first item of playlist, when controller does not support playlist playback.~~
+ ~~Feedback to user when casting YouTube playlists (mass-adding videos).~~
+ ~~Warnings to users, trying to use YouTube functionality with audio devices.~~
+ ~~Tests are of course broken.~~
+ Other stuff I forgot about.
+ ~~Add docstrings.~~

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -122,7 +122,7 @@ def cast(settings, video_url):
 @click.argument("video_url")
 @click.pass_obj
 def add(settings, video_url):
-    cst, stream = setup_cast(settings["device"], video_url=video_url, prep="app")
+    cst, stream = setup_cast(settings["device"], video_url=video_url, prep="control")
     cst.add(stream.video_id)
 
 
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -232,7 +232,7 @@ def scan():
 def writeconfig(settings):
     try:
         os.mkdir(CONFIG_DIR)
-    except:
+    except:  # noqa
         pass
 
     # Put all the standalone options from the settings into an "options" key.
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -103,7 +103,7 @@ def cast(settings, video_url):
         try:
             cst.play_playlist(stream.playlist)
         except PlaybackError:
-            click.echo("Playlist playback not possible, playing first video.", err=True)
+            click.echo("Warning: Playlist playback not possible, playing first video.", err=True)
             if cst.info_type == "url":
                 cst.play_media_url(stream.first_entry_url)
             elif cst.info_type == "id":
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -92,7 +92,7 @@ def cast(settings, video_url):
 
         thr.setDaemon(True)
         thr.start()
-        cst.play_content(stream.video_url)
+        cst.play_media_url(stream.video_url)
         click.echo("Serving local file, press Ctrl+C when done.")
         while thr.is_alive():
             time.sleep(1)
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -123,6 +123,8 @@ def cast(settings, video_url):
 @click.pass_obj
 def add(settings, video_url):
     cst, stream = setup_cast(settings["device"], video_url=video_url, prep="control")
+    if cst.name == "default" or cst.name != stream.extractor:
+        raise CattCliError("This url cannot be added to the queue.")
     cst.add(stream.video_id)
 
 
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 import os
 import time
-
 from threading import Thread
+
 try:
     import ConfigParser as configparser
 except ImportError:
@@ -11,11 +11,11 @@ except ImportError:
 import click
 
 from .controllers import (
+    Cache,
     get_chromecast,
     get_chromecasts,
-    setup_cast,
-    Cache,
-    PlaybackError
+    PlaybackError,
+    setup_cast
 )
 from .http_server import serve_file
 
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -5,19 +5,18 @@ import time
 from threading import Thread
 try:
     import ConfigParser as configparser
-except:
+except ImportError:
     import configparser
 
 import click
 
 from .controllers import (
-    Cache,
-    CastController,
     get_chromecast,
-    get_chromecasts
+    get_chromecasts,
+    setup_cast,
+    Cache
 )
 from .http_server import serve_file
-from .stream_info import StreamInfo
 
 
 CONFIG_DIR = click.get_app_dir("catt")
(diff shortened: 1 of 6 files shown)


catt/controllers.py
```diff
@@ -35,8 +41,37 @@ def get_chromecast(device_name):
...
+        try:
+            if stream.is_local_file:
+                raise ValueError
+            app_id = APP_IDS[stream.extractor]
+        except (KeyError, ValueError):
+            app_id = DEFAULT_APP_ID
+    else:
+        stream = None
+        app_id = cast.app_id
+
```
theychx: I know this is a bit creative, but it seemed like the least weird solution. At one point I considered putting the `stream` object inside the controller object, but decided against it, as I think this solution makes the `CastController` classes themselves more concise. I'd also like to preserve the explicit nature of current `cli.py`.

catt/controllers.py
```diff
@@ -35,8 +41,37 @@ def get_chromecast(device_name):
...
+    if app_id == APP_IDS["youtube"] and cast.cast_type != "audio":
+        controller = YoutubeCastController(cast, "YouTube", APP_IDS["youtube"], prep=prep)
+    else:
+        controller = DefaultCastController(cast, "default", DEFAULT_APP_ID, prep=prep)
+    return (controller, stream) if stream else controller
```
theychx: I know this is a bit creative, but it seemed like the least weird solution. At one point I considered putting the `stream` object inside the controller object, but decided against it, as I think this solution makes the `CastController` classes themselves more concise. I'd also like to preserve the explicit nature of current `cli.py`.

skorokithakis:
Hey, sorry for the delay, will review now. Do I need to wait before the checks are fixed to review?

theychx:
> Do I need to wait before the checks are fixed to review?
No, I wanted your opinion on the overall ideas before completing the PR.

skorokithakis:
Alright great, will post in a bit then, thank you.

catt/cli.py
```diff
@@ -99,34 +95,16 @@ def cast(settings, video_url):
         click.echo("Serving local file, press Ctrl+C when done.")
         while thr.is_alive():
             time.sleep(1)
-    # Google blocks Chromecast Audio devices from running the YouTube app.
-    elif stream.is_youtube_video and cc_type != "audio":
-        click.echo("Casting YouTube video %s..." % stream.video_id)
+
+    elif stream.is_video and cst.name != "default":
```
skorokithakis: Why are we checking for `cst.name != "default"` here? This repeated check looks like it'd be better split off, perhaps?

skorokithakis: Also there should probably be a comment explaining this, as I'm sure others will have the same question.

catt/controllers.py
```diff
@@ -35,8 +41,35 @@ def get_chromecast(device_name):
         return devices[0]
 
 
-def human_time(seconds):
-    return time.strftime("%H:%M:%S", time.gmtime(seconds))
+def setup_cast(device_name, video_url=None, prep=None):
```
skorokithakis: Can you add a docstring explaining what this function does and the arguments?

catt/cli.py
theychx: Yes, this is not done. I haven't yet made up my mind on how to solve this.

catt/controllers.py
theychx: Will do.

catt/cli.py
skorokithakis: Maybe move much of this code to a `play_stream` method that will be more explicit about its actions?

catt/controllers.py
skorokithakis: Hmm, yeah. It makes sense to me, but I'm wondering whether this is a "controller" matter or an "app" matter, ie whether we should have different application classes that get passed to the controller, rather than different controller classes.

E.g. `controller = DefaultCastController(cast, "default", DEFAULT_APP_ID, prep=prep, app=YoutubeApp())`. I'm afraid I don't remember enough about the architecture of the Chromecast to definitely say whether this is a better approach, I'm just wondering whether the catt controller should be decoupled from the code that controls app that's running on the Chromecast.

I'll defer to you on this, as your knowledge is more current.

catt/controllers.py
```diff
@@ -104,91 +137,65 @@ def clear(self):
...
+            self.cast.register_handler(self._controller)
+        except AttributeError:
+            self._controller = self.cast.media_controller
+
+        if prep == "app":
+            self._prep_app()
+        elif prep == "control":
+            self._prep_control()
+
+    def _prep_app(self):
```
skorokithakis: These methods probably need docstrings.

theychx: Yep. I'll add them along with the one for ```setup_cast```.

skorokithakis:
Overall, I prefer your approach much more to the current approach. It's much more decoupled, as the controller for each app is in its own class, and can override code as it needs. I'm missing the high-level overview of how this relates to apps, is the YouTube app controlled from the controller as usual, except some of the behaviour changes? Can you summarize how the Chromecast does this in one or two sentences?

catt/controllers.py
theychx: > I'm just wondering whether the catt controller should be decoupled from the code that controls app that's running on the Chromecast.

I'm not sure I understand what you are saying. My main goal was to throw uniform controller objects back to the cli, regardless of what app the user wants to use/control. How can I achieve that if controller and app-control is still decoupled?

skorokithakis: I meant that the controller will accept its app-specific code as an argument, and the main Controller class will be app-agnostic. As I said, though, I don't really have an intuition of the high-level app flow on the Chromecast, so I'm probably way off. Can you explain a bit how e.g. the YouTube app differs from the default video controller app?

theychx:
There is no real behaviour change, with regards to how we use the YouTube controller.

When the user wants to do something that necessitates the use of a custom app, we initiate/manipulate that app with the custom controller. And as the cc responds to the generic google namespace (i.e. ```cast.media_controller```) regardless of what app it is running, it makes sense to couple the custom controller functionality (playback, playlist manipulation) with the generic ditto (playback control).

skorokithakis:
Right, I see. In that case, your approach makes sense to me, thanks.

theychx:
DIAL connections to my cc audio from pychromecast stopped working (probably because of a cc firmware upgrade). So I'm going to put this on hold, until I figure out whats going on.

theychx:
...............and we're back online. This probably needs a bit more testing though.
```setup_cast``` got a bit involved, but with the benefit of everything else being more straightforward (IMHO).

catt/controllers.py
skorokithakis: Still no docstring though :P

skorokithakis:
I like where this is going, just have a few minor issues with missing docstrings and failing tests. I will review more thoroughly when these are fixed, too.

catt/controllers.py
theychx: I edited the todo list :P

theychx:
Fair enough. A few days of testing this in my setup and then I'll wrap this up (promise).

theychx:
Done

PR #58, opened 2018-02-03
Fix youtube add

theychx:
A few minor fixes for yt queue. 
The crash would occur if the user where to interrupt the yt-app while ```catt``` is trying to add several videos to the queue. 
Listener event flags are now set independently, as this is more correct and might otherwise lockup ```catt.add``` if interrupted while waiting for ```listener.not_buffering``` to be set (I honestly don't remember why I made ```not_buffering``` dependant on ```app_ready```).


skorokithakis:
Thank you!

PR #60, opened 2018-02-06
Add ability to receive path/url from stdin

theychx:
Something I've wanted/needed for a long time, but were too lazy to implement.


skorokithakis:
Very useful, thanks!

PR #62, opened 2018-02-08
Fix for StreamInfo.init

theychx:
Some playlist extractors (like ```bandcamp:album```) does not provide an ```id``` key with entries, so we need to tolerate that. Not really a problem, as ```_playlist_items``` only is needed by custom controllers.

### DIFF
--- a/catt/stream_info.py
+++ b/catt/stream_info.py
@@ -49,7 +49,7 @@ class StreamInfo:
             if self.is_playlist:
                 items = list(self._preinfo["entries"])
                 self._first_entry_info = self._get_stream_info(items[0])
-                # Some playlist extractors does not provide an id key with entries,
+                # Some playlist extractors do not provide an id key with entries,
                 # so we need to tolerate that. _playlist_items is only required by
                 # custom controllers anyway.
                 try:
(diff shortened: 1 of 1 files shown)


catt/stream_info.py
```diff
@@ -49,7 +49,10 @@ def __init__(self, video_url, model=None, host=None):
             if self.is_playlist:
                 items = list(self._preinfo["entries"])
                 self._first_entry_info = self._get_stream_info(items[0])
-                self._playlist_items = [item["id"] for item in items]
+                try:
+                    self._playlist_items = [item["id"] for item in items]
+                except KeyError:
```
skorokithakis: Can you add a comment here, explaining why you're doing this? You're building Chesterton's fence here.

theychx: Sure.

skorokithakis: Great, I'm cool to merge after that, thanks!

theychx:
I'd like to fix that right away. Do you mind re-reviewing?

skorokithakis:
No problem, just push the change. I'll be going to sleep pretty soon though.

theychx:
Cool. And thank you for the interesting reference. :thinking:

skorokithakis:
Any time, man. I bust it out a lot when people are writing code others will be scratching their heads over in a month.

PR #63, opened 2018-02-09
Make info command more useful

theychx:
Values for "volume_level" and "volume_muted" in ```cast.media_controller.status``` are useless (always the same). We now add the data from ```cast.status``` which contain the same volume keys, but with the corresponding values containing actual state.


skorokithakis:
Thank you! Commenting game :100:

PR #66, opened 2018-02-10
Cleanup n freeze

theychx:
This should be the last PR for now.:sweat_smile:

+ I've made ```CastController.cast``` private, now that ```cli.cast``` only needs the cc name.
+ Made ```CastController.skip``` a bit prettier, no real change.
+ Freezed PyChromecast requirement, now that their next release will break Py2 support.
+ Minor change in how we display volume in ```status```, because of weird behaviour from PyChromecast 
   with regards to this (try ```catt info``` after you've adjusted volume).

I hope you don't mind me rolling this up in one PR, but I wanted to move on.

PS: I've temporarily ditched my efforts to implement a verbose option, as it breaks tests, and I honestly can't figure out how to fix them.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -96,11 +96,10 @@ def write_config(settings):
 @click.pass_obj
 def cast(settings, video_url):
     cst, stream = setup_cast(settings["device"], video_url=video_url, prep="app")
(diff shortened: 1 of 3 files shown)


skorokithakis:
Looks great, thank you!

PR #69, opened 2018-02-22
Misc improvements/fixes for final Python 2 release

### LINKED ISSUE #68: Latest version won't execute via PHP
Hey,  

I've been using Cast All The Things in conjunction with a php script via IFTTT and haven't had any issues with it **until** upgrading to v0.5.3, which for some reason cannot seem to work when being called from an elevated script via php.   

The previous version works fine however, so I'm sticking with it for now, but just thought I'd mention it as I'm somewhat stumped as to why it no longer works. (I'm assuming it is some sort of permissions issue or maybe a path related issue or something like that, but as I mentioned, it runs the script as an elevated user and the previous version has no problems)

Let me know if you have any insight :)   Thanks for a great script.

theychx:
Hello!

A few improvements/fixes:
+ Further scrutinize input in ```cli.add``` so we fail instead of crashing, when f.ex. the user tries to add a youtube playlist url.
+ A few special kinds of urls (like yt channel urls) are now re-processed, so those can be played as yt playlist urls.
+ Fixes #68 
+ Proper volume display in python 2

I've put this branch through my testing scheme in BOTH :smile: py2 and 3. I somehow managed to convince myself that I had already done that, but apparently I didn't.

The ```2.0.0``` release of pychromecast (without py2 support) is now out FYI.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -139,8 +139,8 @@ def cast(settings, video_url):
 @click.pass_obj
 def add(settings, video_url):
     cst, stream = setup_cast(settings["device"], video_url=video_url, prep="control")
-    if not stream.is_video:
-        raise CattCliError("Only video urls can be added to the queue.")
+    if (cst.name != "default" and cst.name != stream.extractor) or not stream.is_video:
+        raise CattCliError("This url cannot be added to the queue.")
     cst.add(stream.video_id)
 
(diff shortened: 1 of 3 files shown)


skorokithakis:
Great, thank you! Will release now.

Issue #57, opened 2018-02-03
Pychromecast is dropping suppport for Python 2


theychx:
Howdy.
Pychromecast will no longer support python 2, as of next release (because python-zeroconf will drop python 2 support).
I have a slew of minor PR's coming up after the controllers refactor PR (which is ready, by the way :smiley: ) has been merged. After that I humbly suggest that we make python 3.5 the min. requirement for ```catt```, as this is now available for all major OS'es (the new min. requirement for Pychromecast will be 3.4, I'm just more keen on 3.5).

skorokithakis:
Hello!  
I'm generally loath to remove support for things. If Pychromecast removes support for 2, there's not much we can do, but I generally like erring on the site of being compatible. If there's a version of 3 we need, then by all means, we should use it, but we shouldn't go around turning everything into f-strings or whatever just because.

Other than that, I generally agree with you (I'm reviewing #55 now, by the way), although I would like to keep support for 3.4 as my Raspbian doesn't have 3.5 yet and I can imagine many people using catt on a Raspberry Pi.

theychx:
Min. 3.4 is cool, although raspbian stretch has 3.5, so I'm not sure I understand that argument. But as long as we commit to no longer care about what might not work for <3.4 then I'm happy (no more try except clauses around import statements etc.).

skorokithakis:
(It may have been a while since I last upgraded my raspi)

theychx:
@skorokithakis Please consider making a final python2 compatible release, so we can move on. As for #73, I'm honestly reluctant to invest time in fixing this myself, as I plan to completely rework the server script (I want ```catt``` to spawn an independent process that automatically dies when the cc has stopped/finished playing). If this pans out, I also intend to extend this further, by implementing #46.

skorokithakis:
@theychx Certainly, I can release now. We're sorted for all the py2 compatible things we wanted to release, right?

About #73, I've been meaning to retest, but I'm kind of blocked because `catt` can't locate my Chromecast. I'll debug that when I have some time and try to fix it.

skorokithakis:
Released `0.5.5` and added README information.

theychx:
> We're sorted for all the py2 compatible things we wanted to release, right?

Yep. I'll make a PR that removes py2 compat code after you've made the release.

PR #75, opened 2018-03-17
Remove py2 compat code

theychx:
As discussed in #57 (CI failed because of weird 403's from vimeo, thus the junk commits).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
 
 install: "pip install -r requirements_dev.txt"
 
-# command to run tests, e.g. python setup.py test
 script:
   - python setup.py test
   - flake8 catt
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: python
 python:
-  - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 install: "pip install -r requirements_dev.txt"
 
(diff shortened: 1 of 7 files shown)


theychx:
Wow, this is annoying. I'll fix the tests, so we don't rely on Vimeo.

skorokithakis:
Fantastic, thanks! Semver dictates that the next release should be 1.0.0, so I guess we better do that. Is there anything else you need to add before then? Perhaps the local-serving rework?

theychx:
> Semver dictates that the next release should be 1.0.0

Could you elaborate on that? Seems to me that we're still not quite there (I did read https://semver.org/#how-do-i-know-when-to-release-100).

> Is there anything else you need to add before then? Perhaps the local-serving rework?

Yep, definitely the local-serving rework. I'd also like to add some form of save/restore mechanism, so users can save media location and seek position on a device & app basis, and then recall it later at their leisure.

theychx:
.....and subtitles support          ...........and a option to force the use of the default controller, in case a custom controller stops working.

skorokithakis:
Oh yeah, you're right about semver, I was thinking of "Major version MUST be incremented if any backwards incompatible changes are introduced to the public API" and forgot that 0.x.y is basically "anything goes". Never mind that comment, then!

theychx:
Ok. But it might be worth considering to do the local-serving rework and the force default controller option, and then release. I'm thinking that it is perhaps a bit annoying to python3 users, if the latest version has pins on PyChromecast/zeroconf (I don't mind personally).

skorokithakis:
Yeah, I agree with that. Hopefully I can fix this crashing bug that's been plaguing me in that release as well.

PR #76, opened 2018-03-19
Add force default option to cast command

theychx:
as discussed

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -90,7 +90,7 @@ def write_config(settings):
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url", callback=process_url)
 @click.option("-f", "--force-default", is_flag=True,
-              help="Force use of the default controller.")
+              help="Force use of the default Chromecast app (use if a custom app doesn't work).")
 @click.pass_obj
 def cast(settings, video_url, force_default):
     cst, stream = setup_cast(settings["device"], video_url=video_url,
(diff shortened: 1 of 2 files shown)


catt/cli.py
```diff
@@ -89,9 +89,12 @@ def write_config(settings):
 
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url", callback=process_url)
+@click.option("-f", "--force-default", is_flag=True,
+              help="Force use of the default controller.")
```
skorokithakis: Can you explain a bit more what this does? End users probably won't know what a controller is.

theychx: Fair point, I considered using "default media controller" instead, but figured that that wouldn't be any clearer. Does such elaboration belong in cli help texts?. I can't think of a way to explain this in a few words. Any ideas?

skorokithakis: Hmm, yeah, good point. I can't find a way to summarize it either. Maybe if it's in terms of "what will happen"? What's a reason that would make someone want to use the default controller?

theychx: "default chromecast app", maybe?

skorokithakis: Yeah, that sounds better. Still, *why* would someone use that option? What's your use case for adding it? We can say something like "(use this if X doesn't work)".

theychx: > What's a reason that would make someone want to use the default controller?

If a custom controller is broken.

theychx: fixed

PR #78, opened 2018-04-06
Make seek/status commands more useful with weird stream data

theychx:
(Yes, I'm procrastinating. :innocent:)

Recently my chromecast has begun reporting a duration of zero, with some content (that is seekable) from a local streaming outfit, meaning that `status` is not useful with these.
I taken this opportunity to make ```catt``` fail when the user tries to seek content, where it makes no sense.


skorokithakis:
Looks great, as always!

PR #79, opened 2018-04-07
Simplify cache

theychx:
Hi
The way that ```catt``` works now, it doesn't seem useful to be able to set an arbitrary location for the cache. Likewise, with the cache living in ```/tmp``` and bad ip entries automatically being corrected, it also does not seem useful for the cache to have a fixed lifespan.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -134,28 +134,17 @@ class PlaybackError(Exception):
 
 
 class Cache:
-    def __init__(self, duration=3 * 24 * 3600,
-                 cache_dir=os.path.join(tempfile.gettempdir(), "catt_cache")):
-        self.cache_dir = cache_dir
+    def __init__(self):
+        self.cache_dir = os.path.join(tempfile.gettempdir(), "catt_cache")
         try:
-            os.mkdir(cache_dir)
+            os.mkdir(self.cache_dir)
         except FileExistsError:
             pass
+        self.cache_filename = os.path.join(self.cache_dir, "chromecast_hosts")
 
-        self.cache_filename = os.path.join(cache_dir, "chromecast_hosts")
-
-        if os.path.exists(self.cache_filename):
-            if os.path.getctime(self.cache_filename) + duration < time.time():
-                self._initialize_cache()
-        else:
-            self._initialize_cache()
-
-    def _initialize_cache(self):
-        data = {}
-        devices = pychromecast.get_chromecasts()
-        for device in devices:
-            data[device.name] = device.host
-        self._write_cache(data)
+        if not os.path.exists(self.cache_filename):
+            devices = pychromecast.get_chromecasts()
+            self._write_cache({d.name: d.host for d in devices})
 
     def _read_cache(self):
         with open(self.cache_filename, "r") as cache:
@@ -167,7 +156,6 @@ class Cache:
 
     def get(self, name):
         data = self._read_cache()
-
         # In the case that cache has been initialized with no cc's on the
         # network, we need to ensure auto-discovery.
         if not data:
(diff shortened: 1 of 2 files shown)


skorokithakis:
Sounds good to me. Since we're 3-only now, we can also use `pathlib` for this, which is much better. I'll review now.

theychx:
So you want the switch to ```pathlib``` in another PR, right?

skorokithakis:
Whatever you want, I was just remarking that it'd be easier to work with. This is fine too.

theychx:
Ok, I'll change it right away then, in this PR.

skorokithakis:
Great! There's a nice thing where `mkdir` can accept an argument to not produce an error if the dir exists, so no need for error handling there.

theychx:
> There's a nice thing where mkdir can accept an argument to not produce an error if the dir exists

This feature was added in py3.5. :smiley:

skorokithakis:
Damnit!

catt/cli.py
```diff
@@ -16,8 +16,8 @@
 from .http_server import serve_file
 
 
-CONFIG_DIR = click.get_app_dir("catt")
-CONFIG_FILENAME = os.path.join(CONFIG_DIR, "catt.cfg")
+CONFIG_DIR = Path(click.get_app_dir("catt"))
+CONFIG_FILE = Path(CONFIG_DIR, "catt.cfg")
```
skorokithakis: Btw you can just do `CONFIG_FILE = CONFIG_DIR / "catt.cfg"`.

catt/cli.py
```diff
@@ -263,7 +263,7 @@ def writeconfig(settings):
         for option, value in options.items():
             config.set(section, option, value)
 
-    with open(CONFIG_FILENAME, "w") as configfile:
+    with open(CONFIG_FILE, "w") as configfile:
```
skorokithakis: `with CONFIG_FILE.open("w") as configfile:`

catt/controllers.py
```diff
@@ -135,23 +134,23 @@ class PlaybackError(Exception):
...
     def __init__(self):
-        self.cache_dir = os.path.join(tempfile.gettempdir(), "catt_cache")
+        self.cache_dir = Path(tempfile.gettempdir(), "catt_cache")
         try:
-            os.mkdir(self.cache_dir)
+            self.cache_dir.mkdir()
         except FileExistsError:
             pass
-        self.cache_filename = os.path.join(self.cache_dir, "chromecast_hosts")
+        self.cache_file = Path(self.cache_dir, "chromecast_hosts")
```
skorokithakis: `self.cache_dir / "chromecast_hosts"`

catt/controllers.py
```diff
@@ -135,23 +134,23 @@ class PlaybackError(Exception):
...
 
-        if not os.path.exists(self.cache_filename):
+        if not self.cache_file.exists():
             devices = pychromecast.get_chromecasts()
             self._write_cache({d.name: d.host for d in devices})
 
     def _read_cache(self):
-        with open(self.cache_filename, "r") as cache:
+        with open(self.cache_file, "r") as cache:
```
skorokithakis: `self.cache_file.open()`

skorokithakis:
I've added some better ways to write some of the lines (using `pathlib` more), feel free to address them or merge as-is, they're only minor suggestions.

catt/controllers.py
theychx: Not a fan of this syntax, but I like your other suggestions, so I'll change that.

skorokithakis: Really? Damn, the slash is half the reason I liked pathlib in the first place 😛

PR #81, opened 2018-04-07
Saner empty playlist check

theychx:
I've moved the check to ```cli.cast```, as that makes more sense. Otherwise, the ```play_playlist``` method of any custom controller would have to implement the check. While working on this, I came across a bug in ```stream_info``` that would make ```catt``` crash with empty playlists, so I fixed that.


theychx:
> Sorry

Eh?

skorokithakis:
I took a long time approving it.

PR #82, opened 2018-04-08
More pathlib fun

### LINKED ISSUE #73: Local file casting issues
Howdy

Trying to cast this media file on my CC Ultra:
```youtube-dl -f "best[width <=? 3840][height <=? 2160]" https://vimeo.com/225888984```
as a local file, results in this traceback:
```
  File "/home/benno/projects/chromecast/catt/catt/http_server.py", line 40, in do_GET
    self.wfile.write(data)
  File "/usr/lib/python3.6/socketserver.py", line 775, in write
    self._sock.sendall(b)
BrokenPipeError: [Errno 32] Broken pipe
192.168.0.188 - - [27/Feb/2018 12:43:54] "GET / HTTP/1.1" 200 -
  File "/home/benno/projects/chromecast/catt/catt/http_server.py", line 40, in do_GET
    self.wfile.write(data)
  File "/usr/lib/python3.6/socketserver.py", line 775, in write
    self._sock.sendall(b)
ConnectionResetError: [Errno 104] Connection reset by peer
```
Served from a DLNA-server, the file plays fine (I did check transcoding settings).

It would be nice to have this fixed, before we axe py2 support. I tried tinkering a bit with the server module, but to no avail.

Upon further investigation, this does not appear to be bitrate related. The Chromecast and the server script are somehow confusing each other when trying to cast certain types of videos. Please note that all the media files that I have had issues with originated from Vimeo (and that regular casting of Vimeo videos with ```catt``` is also flakey).

### PR DESCRIPTION
**Pull request by theychx**
Howdy
Forgot about ```stream_info``` and ```http_server``` also using ```os```. I've included the local file casting fix (removal of "Accept-Ranges" header).

Looking forward (short-term), I'm honestly more keen on working on the save/restore stuff right now. So if it's ok with you, I'd be cool if you could make a ```0.6.0``` release after this is merged.

Fixes #73

### DIFF
--- a/catt/http_server.py
+++ b/catt/http_server.py
@@ -1,33 +1,33 @@
-import os
 import socketserver
 import time
 import traceback
 from http.server import BaseHTTPRequestHandler
+from pathlib import Path
 
 
 def serve_file(filename, address="", port=45114):
     class FileHandler(BaseHTTPRequestHandler):
         def do_GET(self):  # noqa
             try:
-                file = open(filename, "rb")
-                stat = os.fstat(file.fileno())
-                length = stat.st_size
+                mediapath = Path(filename)
+                length = mediapath.stat().st_size
+                mtime = mediapath.stat().st_mtime
+                mediafile = open(str(mediapath), "rb")
 
                 self.send_response(200)
                 self.send_header("Content-type", "video/mp4")
                 self.send_header("Content-Length", length)
-                self.send_header("Accept-Ranges", "bytes")
                 self.send_header(
                     "Last-Modified",
                     time.strftime(
                         "%a %d %b %Y %H:%M:%S GMT",
-                        time.localtime(os.path.getmtime(filename))
+                        time.localtime(mtime)
                     )
                 )
                 self.end_headers()
 
                 while True:
-                    data = file.read(100 * 1024)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: catt/http_server.py [hunk 1]
```diff
@@ -1,41 +1,41 @@
...
                         "%a %d %b %Y %H:%M:%S GMT",
-                        time.localtime(os.path.getmtime(filename))
+                        time.localtime(mtime)
                     )
                 )
                 self.end_headers()
 
                 while True:
-                    data = file.read(100 * 1024)
+                    data = mediafile.read(100 * 1024)
```
theychx: It does not appear to be possible to do this with ```pathlib```.

skorokithakis: What isn't possible, you mean?

skorokithakis:
Looks good to me! I can release after merge, yes, but I'm in the middle of something now, so do remind me if I forget.

catt/http_server.py
theychx: Partial reading, which is why I used built-in ```open()```.

skorokithakis: Ah, yes, unfortunately the `read()` parameter doesn't take ranges.

PR #85, opened 2018-04-16
Improve kill

theychx:
Just a little tweak so we don't kill runnnig cc sessions on certain type of fails.

..and a fix for a bug from the previous PR.


skorokithakis:
LGTM!

PR #86, opened 2018-04-17
Play random playlist entry / StreamInfo improvements

### LINKED ISSUE #83: Random playlist
Would it be possible to add the ability to play random video off a playlist?

theychx:
Gave me an opportunity to get rid of the lame ```first_entry``` stuff in ```StreamInfo```.

Closes #83

### DIFF
--- a/catt/stream_info.py
+++ b/catt/stream_info.py
@@ -137,7 +137,13 @@ class StreamInfo:
         return self._active_entry.get("thumbnail") if self.is_playlist else None
 
     def set_playlist_entry(self, number):
+        """
+        Must be called with valid entry number
+        before playlist entry properties can be accessed.
+        """
+
         if self.is_playlist:
+            # Some playlist entries needs to be re-processed.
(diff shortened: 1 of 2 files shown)


catt/stream_info.py
```diff
@@ -68,17 +61,11 @@ def is_video(self):
 
     @property
     def is_playlist(self):
-        if not self.is_local_file:
-            return True if "entries" in self._preinfo else False
-        else:
-            return False
+        return True if not self.is_local_file and "entries" in self._preinfo else False
```
skorokithakis:
```python
return not self.is_local_file and "entries" in self._preinfo
```

skorokithakis:
One minor comment, but I can merge anyway.

catt/stream_info.py
theychx: Cheers. Why didn't I think of that?

skorokithakis: Oh it gets me every time too.

PR #87, opened 2018-04-18
Add color to warning messages

theychx:
:art:

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -142,7 +142,8 @@ def cast(settings, video_url, force_default, random_play):
                 cst.play_playlist(stream.playlist_all_ids)
                 return
             except (PlaybackError, ValueError):
-                click.echo("Warning: Playlist playback not possible, playing first video.", err=True)
+                click.secho("Warning: ", fg="red", nl=False)
+                click.echo("Playlist playback not possible, playing first video.", err=True)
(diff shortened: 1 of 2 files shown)


catt/cli.py
```diff
@@ -283,8 +284,8 @@ def save(settings, path):
     if not cst.save_capability or cst.is_streaming_local_file:
         raise CattCliError("Saving state of this kind of content is not supported.")
     elif cst.save_capability == "partial":
-        click.echo("Warning: Please be advised that playlist data will not be saved.",
-                   err=True)
+        click.secho("Warning: ", fg="red", nl=False)
```
skorokithakis: What do you think about putting this in a `warning()` function so we don't repeat this line every time?

theychx: Yeah, I thought about too, but came to the conclusion that with only 3 warnings it would not save any lines of code. But it's not important to me, so if you prefer using a ```warning()``` function, I'll change it, no probelm.

skorokithakis: Hmm, it's not that important to me either, it just saves on duplication so if we want to make it, say, light red in the future, we only have one place to change.

To clarify, I mean something like `warning("Please be advised that playlist data will not be saved.")`.

I'll approve either way, as it's not a deal-breaker for me, just a nice-to-have.

PR #88, opened 2018-04-18
Improve save/restore

theychx:
:floppy_disk:

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -208,12 +208,12 @@ class CastState(CattStore):
     def get_data(self, name):
         try:
             data = self._read_store()
-            save_data = data.get(name)
-            if save_data and set(save_data.keys()) != set(["controller", "data"]):
+            if set(next(iter(data.values())).keys()) != set(["controller", "data"]):
                 raise ValueError
-        except (json.decoder.JSONDecodeError, ValueError):
+        except (json.decoder.JSONDecodeError,
+                ValueError, StopIteration, AttributeError):
             raise StateFileError
-        return save_data
+        return data.get(name)
 
 
 class CastStatusListener:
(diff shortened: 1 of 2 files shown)


catt/controllers.py
```diff
@@ -208,12 +211,12 @@ def __init__(self, state_path):
     def get_data(self, name):
         try:
             data = self._read_store()
-            save_data = data.get(name)
-            if save_data and set(save_data.keys()) != set(["controller", "data"]):
+            if set(next(iter(data.values())).keys()) != set(["controller", "data"]):
```
skorokithakis: This is obscure enough that a comment is warranted, I think. My problem with Python in general is that I never know the type, so I can't reason about things. For example, I know you're looking for the first truthy value here, but I have no idea reading this line what type `data` is. Maybe a comment with what `data` looks like and a comment saying "Get the first truthy element" would clarify things more.

skorokithakis:
Sorry for the delay, I've been pretty busy :/

catt/controllers.py
theychx: LOL, yeah this looks pretty weird. The problem I was trying to solve is that keys in the root level are cc names, so I can't verify those, and so I'm just trying to validate the data associated with whatever cc name. Maybe I'm misunderstanding you, what do you mean by "truthy"?

skorokithakis: Oh sorry, I misunderstood the code. I'm confused as to why you do `next(iter(data.values()))` rather than just `data.values()[0]`?

theychx: ```data.values``` does not do indexing, as far as I can tell.

skorokithakis: Is `data` not just a dictionary? It's returned by `json.read()`, so it should be, no?

theychx: Yep, exactly:
```'dict_values' object does not support indexing```

skorokithakis: Oh bah, I forgot that's a view object in py3... And it doesn't even support iteration, how inconvenient. `list(data.views())[0]` might be more readable, but not by much. Alright, looks good to me then!

PR #92, opened 2018-05-02
Use a more educational example Youtube link

ofek:
@skorokithakis Thanks so much for this tool!


skorokithakis:
The new link better be a Rickroll.

skorokithakis:
I love you.

Issue #46, opened 2017-08-22
Subtitles support


owyongsk:
Would it be possible?

skorokithakis:
I'm not sure, I haven't looked into it, unfortunately.

owyongsk:
Does the ChromeCast api allow subtitles for a web URL?

skorokithakis:
That's a good question, I think so, but I don't know. If Pychromecast has support for subtitles, it would probably be easy to implement.

owyongsk:
Hmm, I managed to add support to another tool here, so it's possible, anyways, I got my use case figured out. Thanks!

https://github.com/owyongsk/stream2chromecast/commit/da7c239fc9d34464ab4c826090775265923ecb61

skorokithakis:
Oh, very interesting! That doesn't seem too hard, I'll see if I can get something implemented quickly (as I need this myself too). Thanks!

owyongsk:
Good luck, I needed this to learn Spanish in my subtitles. Haha.

skorokithakis:
Damn! I'll keep this open, as we do need this. Thanks for the report!

marcosdiez:
@owyongsk I checked your PR and saw that it's quite easy to add subtitles on `catt`. Thanks for the nice hint.

PR #93, opened 2018-05-03
feat: Cast file with local or remote subtitles

marcosdiez:
fixes https://github.com/skorokithakis/catt/issues/46

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,5 @@ install: "pip install -r requirements_dev.txt"
 # command to run tests, e.g. python setup.py test
 script:
   - python setup.py test
+  - flake8 --version
   - flake8 catt
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+
 import configparser
 import os
 import random
@@ -9,6 +10,7 @@ from pathlib import Path
 from threading import Thread
 
 import click
+import requests
 
 from .controllers import (
     Cache,
@@ -120,21 +122,25 @@ def hunt_subtitle(video):
     return None
 
 
+def convert_srt_to_webvtt_helper(content):
+    content = re.sub(r'([\d]+)\,([\d]+)', r'\1.\2', content)
+
+    with tempfile.NamedTemporaryFile(mode='w+b',
+                                     suffix=".vtt",
+                                     delete=False) as vttfile:
+        target_filename = vttfile.name
+        vttfile.write("WEBVTT\n\n".encode())
+        vttfile.write(content.encode())
+        return target_filename
+
+
 def convert_srt_to_webvtt(filename):
     # print("Converting {} to WebVTT".format(filename))
     for possible_encoding in ['utf-8', 'iso-8859-15']:
         try:
             with open(filename, 'r', encoding=possible_encoding) as srtfile:
                 content = srtfile.read()
-                content = re.sub(r'([\d]+)\,([\d]+)', r'\1.\2', content)
-
-                with tempfile.NamedTemporaryFile(mode='w+b',
-                                                 suffix=".vtt",
-                                                 delete=False) as vttfile:
-                    target_filename = vttfile.name
-                    vttfile.write("WEBVTT\n\n".encode())
-                    vttfile.write(content.encode())
-                    return target_filename
+                return convert_srt_to_webvtt_helper(content)
         except UnicodeDecodeError:
             pass
     raise CattCliError("Could not find the proper encoding of {}. Please convert it to utf-8".format(filename))
@@ -145,11 +151,15 @@ def load_subtitle_if_exists(subtitle, dont_load_subtitles_automatically, video, 
         subtitle = hunt_subtitle(video)
     if subtitle is None:
         return None
-    print("Using subtitle {}".format(subtitle))
+    click.echo("Using subtitle {}".format(subtitle))
 
     if "://" in subtitle:
         # it's an URL
-        return subtitle
+        if subtitle.lower().endswith(".srt"):
+            content = requests.get(subtitle).text
+            subtitle = convert_srt_to_webvtt_helper(content)
+        else:
+            return subtitle
 
     if subtitle.lower().endswith(".srt"):
         subtitle = convert_srt_to_webvtt(subtitle)
@@ -193,7 +203,7 @@ def cast(settings, video_url, subtitle, force_default, random_play, dont_load_su
     if stream.is_local_file:
         click.echo("Casting local file %s..." % video_url)
         click.echo("Playing %s on \"%s\"..." % (stream.video_title, cst.cc_name))
-        subtitle_url = load_subtitle_if_exists(subtitle, dont_load_subtitles_automatically, video_url, stream.local_ip, stream.port+1)
+        subtitle_url = load_subtitle_if_exists(subtitle, dont_load_subtitles_automatically, video_url, stream.local_ip, stream.port + 1)
 
         thr = Thread(target=serve_file,
                      args=(video_url, stream.local_ip, stream.port))
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import configparser
+import os
 import random
 import re
 import tempfile
@@ -106,14 +107,16 @@ def write_config(settings):
 
 
 def hunt_subtitle(video):
-    dot_pos = video.rfind(".")
-    if dot_pos < 0:
-        return None
-    naked_video = video[0:dot_pos]
-    for expected_sufix in [".vtt", ".VTT", ".srt", ".SRT"]:
-        new_name = naked_video + expected_sufix
-        if Path(new_name).is_file():
-            return new_name
+    """"Searches for subtitles in the current folder"""
+    video_path = Path(video)
+    video_path_stem_lower = video_path.stem.lower()
+    for entry in os.listdir(video_path.parent):
+        entry_path = Path(entry)
+        if entry_path.is_dir():
+            continue
+        if entry_path.stem.lower().startswith(video_path_stem_lower) and \
+                entry_path.suffix.lower() in [".vtt", ".srt"]:
+            return entry
     return None
 
 
@@ -139,8 +142,8 @@ def convert_srt_to_webvtt(filename):
 
 def load_subtitle_if_exists(subtitle, video, local_ip, port):
     subtitle = subtitle or hunt_subtitle(video)
-        if subtitle is None:
-            return None
+    if subtitle is None:
+        return None
     print("Using subtitle {}".format(subtitle))
 
     if "://" in subtitle:
@@ -187,7 +190,7 @@ def cast(settings, video_url, subtitle, force_default, random_play):
     if stream.is_local_file:
         click.echo("Casting local file %s..." % video_url)
         click.echo("Playing %s on \"%s\"..." % (stream.video_title, cst.cc_name))
-        subtitle_url = load_subtitle_if_exists(subtitle, video_url, stream.local_ip, stream.port)
+        subtitle_url = load_subtitle_if_exists(subtitle, video_url, stream.local_ip, stream.port+1)
 
         thr = Thread(target=serve_file,
                      args=(video_url, stream.local_ip, stream.port))
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -140,8 +140,9 @@ def convert_srt_to_webvtt(filename):
     raise CattCliError("Could not find the proper encoding of {}. Please convert it to utf-8".format(filename))
 
 
-def load_subtitle_if_exists(subtitle, video, local_ip, port):
-    subtitle = subtitle or hunt_subtitle(video)
+def load_subtitle_if_exists(subtitle, dont_load_subtitles_automatically, video, local_ip, port):
+    if subtitle is None and not dont_load_subtitles_automatically:
+        subtitle = hunt_subtitle(video)
     if subtitle is None:
         return None
     print("Using subtitle {}".format(subtitle))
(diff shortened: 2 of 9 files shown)


skorokithakis:
Daaamn, this is amazing, thank you! @theychx, do you want to review this with me?

theychx:
Yep.

catt/cli.py
```diff
@@ -103,28 +105,98 @@ def write_config(settings):
         raise CattCliError("No device specified.")
 
 
+def hunt_subtitle(video):
+    dot_pos = video.rfind(".")
```
skorokithakis: It would be nice to use Pathlib here, and also to write a more robust algorithm. My subtitle names aren't always `videofile.srt`, sometimes they are `videofile.en.srt` and `videofile.el.srt`, so it would be good to do some globbing and fetch the first file alphabetically, if nothing is selected.

catt/cli.py
```diff
@@ -103,28 +108,109 @@ def write_config(settings):
...
+    """"Searches for subtitles in the current folder"""
+    video_path = Path(video)
+    video_path_stem_lower = video_path.stem.lower()
+    for entry in os.listdir(video_path.parent):
+        entry_path = Path(entry)
+        if entry_path.is_dir():
+            continue
+        if entry_path.stem.lower().startswith(video_path_stem_lower) and \
+                entry_path.suffix.lower() in [".vtt", ".srt"]:
+            return entry
```
theychx: This is not working, as you're not returning the absolute path. You could do this:
```for entry_path in video_path.parent.iterdir()```
(and thus, no longer need ```os```)
And then let the function return ```str(entry_path.resolve())```

File: catt/cli.py [hunk 3]
```diff
@@ -103,28 +108,109 @@ def write_config(settings):
...
-def cast(settings, video_url, force_default, random_play):
+def cast(settings, video_url, subtitle, force_default, random_play, dont_load_subtitles_automatically):
     controller = "default" if force_default else None
     cst, stream = setup_cast(settings["device"], video_url=video_url,
                              prep="app", controller=controller)
 
     if stream.is_local_file:
         click.echo("Casting local file %s..." % video_url)
         click.echo("Playing %s on \"%s\"..." % (stream.video_title, cst.cc_name))
+        subtitle_url = load_subtitle_if_exists(subtitle, dont_load_subtitles_automatically, video_url, stream.local_ip, stream.port + 1)
```
theychx: I would prefer if the flag-handling logic was here, instead of the flag being passed on to the function. Like:
```
if subtitle is None and dont_load_subtitles_automatically:
    subtitle_url = None
else:
    subtitle_url = load_subtitle_if_exists(subtitle, video_url, stream.local_ip, stream.port + 1)
```

marcosdiez:
Sorry for the delay. 
I was annoyed `catt` was not working on WSL and it took me a while to fix that.
It turns out the problem was in `python-zeroconf` ( https://github.com/jstasiak/python-zeroconf/pull/130 )  so as soon as that get merged into `pychromecast` I'll be back here and ask you for another release :)

marcosdiez:
By the way, @theychx I don't know what's wrong with `travis` and `flake8` today... it's not my fault. sorry for that. I had the same issue on the python zeroconf PR :(

skorokithakis:
I've had issues with them being flaky as well. Usually just rerunning the test works.

marcosdiez:
@skorokithakis I rebased to make sure travis would run again and it failed again. My other PR also failed again, for the same reason...

any ideas ?

theychx:
@marcosdiez The travis fails does not seem to be caused by anything in your code. I have not idea whats going on.

skorokithakis:
I have fixed the Travis failure. Let me see if I can update this branch with the fix.

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
         raise CattCliError("No device specified.")
 
 
+def hunt_subtitle(video):
+    """"Searches for subtitles in the current folder"""
+    video_path = Path(video)
+    video_path_stem_lower = video_path.stem.lower()
+    for entry in video_path.parent.iterdir():
+        entry_path = Path(entry)
```
skorokithakis: Is this necessary? Aren't the returned objects paths already?

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
+        if entry_path.is_dir():
+            continue
+        if entry_path.stem.lower().startswith(video_path_stem_lower) and \
+                entry_path.suffix.lower() in [".vtt", ".srt"]:
+            return str(entry_path.resolve())
+    return None
+
+
+def convert_srt_to_webvtt_helper(content):
+    content = re.sub(r'([\d]+)\,([\d]+)', r'\1.\2', content)
```
skorokithakis: This seems a bit too general, it will catch things like internationalized numbers in the subtitles and mistakenly convert them. Can we make it more specific?

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
+            with open(filename, 'r', encoding=possible_encoding) as srtfile:
+                content = srtfile.read()
+                return convert_srt_to_webvtt_helper(content)
+        except UnicodeDecodeError:
+            pass
+    raise CattCliError("Could not find the proper encoding of {}. Please convert it to utf-8".format(filename))
+
+
+def load_subtitle_if_exists(subtitle, video, local_ip, port):
+    subtitle = subtitle or hunt_subtitle(video)
```
skorokithakis: Sorry to change this again, but the ternary `if` is more readable here:

```python
subtitle = subtitle if subtitle else hunt_subtitle(video)
```

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
+    if "://" in subtitle:
+        # it's an URL
+        if subtitle.lower().endswith(".srt"):
+            content = requests.get(subtitle).text
+            subtitle = convert_srt_to_webvtt_helper(content)
+        else:
+            return subtitle
+
+    if subtitle.lower().endswith(".srt"):
+        subtitle = convert_srt_to_webvtt(subtitle)
```
skorokithakis: Shouldn't this give an error unless the extension is `.srt` or `.vtt` instead? We wouldn't want to continue with any arbitrary file.

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
+
+    if not Path(value).is_file():
+        raise CattCliError("Subtitle file [{}] does not exist".format(value))
+
+    return value
+
+
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url", callback=process_url)
+@click.argument("subtitle", required=False, callback=process_subtitle)
```
skorokithakis: I think this would be better served as an option, rather than as an argument.

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
 @click.option("-f", "--force-default", is_flag=True,
               help="Force use of the default Chromecast app (use if a custom app doesn't work).")
 @click.option("-r", "--random-play", is_flag=True,
               help="Play random item from playlist, if applicable.")
+@click.option("--dont-load-subtitles-automatically", is_flag=True, default=False,
```
skorokithakis: Maybe this should be renamed to `--no-subs`? It's kind of long right now.

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
+              help="Don't try to load subtitles automatically from the local folder.")
 @click.pass_obj
-def cast(settings, video_url, force_default, random_play):
+def cast(settings, video_url, subtitle, force_default, random_play, dont_load_subtitles_automatically):
```
skorokithakis: Having `subtitle` be an option would work better here as well, as I don't like changing function signatures that drastically. However, we don't have default arguments for this, so it'd break anyway.

skorokithakis:
Thanks again for this PR, @marcosdiez, this is extremely useful. I'll release 0.7.0 as soon as we merge.

catt/cli.py
```diff
@@ -103,28 +107,111 @@ def write_config(settings):
...
     controller = "default" if force_default else None
     cst, stream = setup_cast(settings["device"], video_url=video_url,
                              prep="app", controller=controller)
 
     if stream.is_local_file:
         click.echo("Casting local file %s..." % video_url)
         click.echo("Playing %s on \"%s\"..." % (stream.video_title, cst.cc_name))
+        if subtitle is None and dont_load_subtitles_automatically:
```
skorokithakis: Is there any reason we can't cast remote files with local subtitles as well? It's going to be a bit more tricky, but if it's easy to do, why not?

theychx: This would not work with save/restore, as we would face the same problem as with save/restore for local files. And if remote files + local subtitles was implemented, users would probably expect this to work.

skorokithakis: Ah, alright.

theychx:
@skorokithakis 
> I'll release 0.7.0 as soon as we merge

Please wait a few days before releasing.

catt/cli.py
theychx: I agree, but that would also change the applicability of ```--dont-load-subtitles-automatically```. Because if this was an option, you might as well just say that ```--subtitle``` without any argument means "try to automatically load subtitle".

skorokithakis: Yes, exactly. And `--no-subtitle` would mean "don't load, even automatically".

skorokithakis: The problem is that there are three states, file/auto/none. So we need `--subtitle`/nothing/`--no-subtitle`.

theychx: Not specifying ```--subtitle``` means none.
Specifying ```--subtitle``` without an argument means auto.
Specifying ```--subtitle <subtitle>``` means file.
What am I missing?

catt/cli.py
skorokithakis: Oh, I misread "subtitle without any argument". I think loading the associated sub should be the default behaviour, without needing an argument...

skorokithakis: I.e. without needing to specify `--subtitle`, I mean.

catt/cli.py
theychx: So:
Nothing specified means "auto".
Specifying ```--subtitle <subtitle>``` means "file".
Specifying ```--no-subtitle``` means "none".

Having given this a bit of thought, I can see that this is probably the behaviour most users would expect.

catt/cli.py
marcosdiez: According to https://developers.google.com/cast/docs/media , Chromecast supports:

    TTML - Timed Text Markup Language
(example, https://gist.github.com/anotherhale/676a72edc84ca3a37c0c )
    WebVTT - Web Video Text Tracks
(example: https://github.com/brenopolanski/html5-video-webvtt-example/blob/master/MIB2-subtitles-pt-BR.vtt
    Line 21 Data Services
example: https://en.wikipedia.org/wiki/EIA-608

I confess I never heard about any of these 3 and I bet nobody will ever use cast with the latter. And XML is a bitch extension, so I think we are fine doing what the user tells us to instead of complaining.

What do you think ?

catt/cli.py
marcosdiez: good, we all agreed on the same thing. I just got somebody visiting me, so I'll continue later.

catt/cli.py
theychx: The problem is that your code is not failing if the specified subtitle file has neither an ".srt" or ".vtt" extension.
(you are right that we should not concern ourselves with the other two obscure formats)

marcosdiez: My point is there may be somebody who would like to use one of the obscure formats and if we arbitrarily forbid it because the extensions we are restricting for restriction's sake.

In the end of the day there is no downside in sending the wrong file to chromecast as an extension, so I think we should not restrict it

marcosdiez:
I added ` "requests>=2.0",` to `setup.py`. I got this version because it's the same on pychromecast, so this way the requirement is there but won't cause an extra download

catt/cli.py
marcosdiez: Also, we don't check the extension for the main files we cast, so we should not bother with the subtitles either.

skorokithakis: Hmm, okay, this makes sense to me, thanks.

.travis.yml
```diff
@@ -9,4 +9,5 @@ install: "pip install -r requirements_dev.txt"
 # command to run tests, e.g. python setup.py test
 script:
   - python setup.py test
+  - flake8 --version
```
skorokithakis: This is probably unnecessary, the version is printed when installing.

catt/http_server.py
```diff
@@ -5,18 +5,30 @@
 from pathlib import Path
 
 
-def serve_file(filename, address="", port=45114):
+def serve_file(filename, address="", port=45114, content_type=None):
     class FileHandler(BaseHTTPRequestHandler):
+
+        def format_size(self, size):
```
skorokithakis: What is this for?

.travis.yml
marcosdiez: I can take it out, but for me it was *very* useful. 
I was doing some other development here and flake was telling me other stuff. If you check it out I have PR called `flake8 is happy again` and another one called `Revert "flake8 is happy again"`, because different flake versions give me different output, so that's the easy way to know what sublibraries `flake8` has. 

I applied the same trick here https://github.com/jstasiak/python-zeroconf/pulls and that's why I am the only one with a passing build. I would keep it. It does not affect the end user and make the devs happier.

catt/http_server.py
marcosdiez: Very useful debugging information.
Now we get this kind of logs:
`10.1.1.3 - - [07/May/2018 11:53:56] "GET /?loaded_from_catt HTTP/1.1" 200 - video/mp4 - 39.80 MB`
`10.1.1.3 - - [07/May/2018 11:53:56] "GET //tmp/tmpmnfqm1im.vtt HTTP/1.1" 200 - text/vtt;charset=utf-8 - 125.97 KB`

I backported from my other PR.

I like knowing how big the file and the mime is useful for the other one anyway.

.travis.yml
skorokithakis: Yeah, that's true, I just generally dislike putting too much stuff in PRs. This is fine though, thanks.

skorokithakis:
Alright, this looks good to me! @theychx, any feedback/objections?

mdiez-modus:
NIIIIIIIIIIIIIIIIIICE !

PR #96, opened 2018-05-05
fix: Specify the correct mimetype when casting local files

marcosdiez:
To make your weekend even more fun, here is a small PR to make `catt` cast images as well :)

yes, I am very optimistic you will soon merge my other PR so I not separating this one, for it will cause conflicts

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -181,6 +181,25 @@ def process_subtitle(ctx, param, value):
     return value
 
 
+def get_mime(path):
+    # source: https://developers.google.com/cast/docs/media
+    extension = Path(path).suffix.lower()
+    extensions = {
+        ".mp4": "video/mp4",
+        ".m4a": "audio/mp4",
+        ".mp3": "audio/mp3",
+        ".mpa": "audio/mpeg",
+        ".webm": "video/webm",
+        ".mkv": "video/x-matroska",
+        ".bmp": "image/bmp",
+        ".jpg": "image/jpeg",
+        ".gif": "image/gif",
+        ".png": "image/png",
+        ".webp": "image/web",
+    }
+    return extensions.get(extension, "video/mp4")
+
+
 @cli.command(short_help="Send a video to a Chromecast for playing.")
 @click.argument("video_url", callback=process_url)
 @click.option("-s", "--subtitle",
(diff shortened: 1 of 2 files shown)


marcosdiez:
@theychx more fun for you! ANother PR for you to code review.

skorokithakis:
Hmm, the problem with having this PR on top of the other is that you may have to redo any changes in the previous one. Hopefully it'll work out, we'll see what git thinks!

marcosdiez:
Don't worry!
the plan is simple:
I'll keep updating the other one until you merge it.

than I'll rebase this one and you can merge this one as well.

this one is incredibly simple: https://github.com/skorokithakis/catt/pull/96/commits/3e01ffa19cf9e1d985fc0b4845cf837524bba510

theychx:
@skorokithakis Is this worth the extra dependency? ```youtube-dl``` only has audio/video extractors (AFAIK).

skorokithakis:
Two extra dependencies, no? `requests` was just not included in the file. Let's get the other one merged first so we can diff this one properly, but I'm not a fan of the extra dependencies either.,

theychx:
```requests``` is pulled in by ```pychromecast```.

marcosdiez:
Thd `Requests` depenency is used by the other PR to download the remote SRT subtitles and convert them to WEBVTT. Since `pychromecast ` uses it, I don't see any point in not using it as well. 

The other dependenxy is `python-magic` that I used because I am lazy.

But in the end of the day the latter is not necessary. All Chromecast supported formats are here : https://developers.google.com/cast/docs/media so I can make a big IF to do the samw thing. Probably only on Monday, though...

theychx:
@marcosdiez The point is, even if you implemented the magic stuff yourself, the only real benefit to merging this, is users gaining the ability to cast local image files. Which I think most users would not expect ```catt``` to do, as the whole point of ```catt``` is to initiate/control audio/video playback.

marcosdiez:
I personally think it's really fun to cast images...

marcosdiez:
I rebased. Now we don't use that mime library. Instead I infer.
The only difference between this PR and the other one is https://github.com/skorokithakis/catt/pull/96/commits/c31394f6364a5b7c5f01ee3a892342e80fe40751

marcosdiez:
rebased. flake8 is happy about it as well!

skorokithakis:
This looks good to me! @theychx?

skorokithakis:
As for the usefulness of this, I think it's more of a "set the correct mimetype" fix, with an added side-effect of allowing casting of audio/image files. I'm not opposed to that, really...

mdiez-modus:
Hey, give me an opinion...

what else could we do with the Chromecast.
I was reading the Google API yesterday and basically we could display any webpage there and also cast random info from the PC to it.

I am wondering if there is a use to that.

Maybe a marquee ? 

I don't know 
any ideas ?
if it's fun I can implement

skorokithakis:
A picture frame would be really fun, maybe we can make a small site that takes a Flickr album or whatever and shows the photos one by one, and the Chromecast can cast that. That'd be a nice screensaver/idle display.

theychx:
Casting of audio files is already working, as chromecast devices do not care about mime-types being correct (for audio/video).

I'm not a fan of latching random functionality onto ```catt```, just for the sake of it.

theychx:
I'm obviously not opposed to setting the correct mime-type.
My concern here is this project loosing its focus. This thread already contains ideas that I think are completely irrelevant to the stated goal of ```catt```.

skorokithakis:
Those wouldn't be implemented in `catt`, it's written in a way that you can easily call it from another script, and the script would just import `catt` and implement the functionality that way. I've always wanted `catt` to be usable as a library as well, even though it's strayed a bit from that lately.

skorokithakis:
Are you okay with merging this PR, by the way?

theychx:
Yeah, it's cool.

skorokithakis:
Thank you, merged!

PR #101, opened 2018-05-11
Allow the content-type to be specified in the cast command

Konubinix:


skorokithakis:
Thank you for your PR! What is the motivation for this? In case we guess the type incorrectly?

catt/cli.py
```diff
@@ -255,7 +259,8 @@ def cast(settings, video_url, subtitle, force_default, random_play, no_subs):
         if cst.info_type == "url":
             cst.play_media_url(stream.playlist_entry_url,
                                title=stream.playlist_entry_title,
-                               thumb=stream.playlist_entry_thumbnail)
+                               thumb=stream.playlist_entry_thumbnail,
+                               content_type = content_type or "video/mp4")
```
skorokithakis: Why not make sure higher up that this exists, and avoid checking again here and farther down?

Konubinix:
The guess works fine in the case of a local file. In the case of a url though, we cannot guess the content type and default to video/mp4. This is the reason why I let the user inform the content type in that situation.

If the content-type is provided in the command line, it sounded sensible that it would override the guessed one.

catt/cli.py
Konubinix: Because I could not find a way to write it more clearly, provided that this check is made only in the url case.

theychx:
@Konubinix Do you have a use case, where this would be a problem with current code?

skorokithakis:
Also, shouldn't we trust the server's mime type in that case?

Konubinix:
@theychx. I encountered the problem when wanting to stream an jpg file. Casting a local image works, since the mimetype is found to be image/jpeg, yet when casting from the url to my nas, the default mimetype video/mp4 makes the chromecast unhappy.

Actually, I mostly stream video/webm files. With the default mimetype of video/mp4, everything works fine, but I think it is better to give the exact mime-type to avoid issues in the future.

Konubinix:
@skorokithakis. I like this idea. That would definitely do it.

skorokithakis:
I guess that's the best we can do, short of issuing a HEAD request to the server and actually reading the header. I'm torn between doing that and just guessing based on the extension, what do you think? Of course, we could also have the latter as a fallback for the former.

theychx:
With the mime-type only being important when casting images, I really think that we should go with the simplest possible solution.

skorokithakis:
Ah, if video files work regardless of mime type, yes, let's just go with the easy solution, agreed.

Konubinix:
The funny thing is that the youtube_dl library already performs a HEAD request to get the content-type and then deduce the extension (see https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/generic.py#L2198). It would be awesome if it returned the content-type as well as the extension. The code of youtube_dl is big and complex enough for me not to try to find out why the content-type is not returned. It probably has good reasons not to do so.

I propose the following:
* expose the ext part of _info in StreamInfo in a property with the name ext
* make get_mime work when only the extension is given, so that get_mime(".mp4") returns "video/mp4"
* use get_mime("." + stream.ext) in the cast command to guess the content-type

Still, I cannot find a way to factorize the calls to get_mime("." + stream.ext) that will be easy to read. I still fall on content_type=get_mime("." + stream.ext).

I will change the pull request accordingly to the above description.

skorokithakis:
Create a property on `stream` called `mime_type` and call `get_mime` there? I'm not crazy about the `"something" + ext` part, though, isn't the entire filename available somewhere in the stream?

Konubinix:
@skorokithakis I took your comments into account

theychx:
I propose the following:
In ```StreamInfo```:
```
@property
def direct_url_mime_type(self):
    if self.is_video and self._info.get("direct"):
        return get_mime(self._info["webpage_url_basename"])
    else:
        return None
```
And then do:
```
content_type=stream.direct_url_mime_type or "video/mp4"
```

theychx:
```is_remote_file``` is the obvious choice here. Cheers.

theychx:
I updated my proposal.

skorokithakis:
I think that makes sense, and I think `get_mime` belongs in `utils` (and should probably be called `guess_mime`). Sounds good to me!

theychx:
@skorokithakis I updated the proposal again, so that mime stuff for both local and remote files are done in ```StreamInfo```. (:face_with_head_bandage: sorry)

skorokithakis:
The updated proposal makes the most sense to me. I wanted to suggest it, but I figured you'd have a good reason not to do the guessing in that method, and I'm not near a computer so I can check. I'm glad this is actually doable, the only change is make would be to name the property `guessed_content_type` or to make it a method called `guess_content_type()`.

Konubinix:
With git grep video | wc -l giving 146 matches. Renaming video to media is not an easy task. I suggest to implement the guessed_content_type first, and dedicate another task to the renaming video -> media.

Is that ok with you if I proposed here the guessed_content_type solution, without renaming video into media?

Konubinix:
I took into account your comments, with the exception of the renaming of video -> media

skorokithakis:
That works fine for me, but we were suggesting renaming `is_video` to `is_remote_file`, which is only used in four places.

skorokithakis:
I'm busy for the next hour or so (debugging something unrelated). If you're free in the mean time, I'd appreciate it, otherwise I'll do it in a bit.

theychx:
Are you merging this, or should I push commits to this PR?

skorokithakis:
You can push here, I figure we might as well include the rename, since this branch is mildly refactory anyway.

theychx:
I screwed up the last commit message. I meant ```guessed_content_type``` not ```guess_mime```.
(having the "video/mp4" fallback in ```play_media_url``` makes things simpler)

I'm going to bed now. We should consider replacing all instances of the word ```video``` with ```media```.

skorokithakis:
It doesn't matter about the commit, we're squashing anyway. Agreed about video and media, PyCharm can do that pretty quickly, and we don't need to do it in this PR. Will merge now, thank you all!

skorokithakis:
By the way, @theychx, will things still work well if we pass `None` as the content-type instead of mp4 when guess_mime can't guess?

theychx:
Yes, there's a "video/mp4" fallback in ```play_media_url```.

skorokithakis:
That will only work when the argument hasn't been passed, though, not when it's `None` (unless I'm mistaken).

theychx:
You're right. Let me fix that.

skorokithakis:
Approved, thanks, everyone. I'll merge as soon as CI passes.

Konubinix:
Thanks for the help :-).

PR #102, opened 2018-05-14
Support to Cast Any Website to Chromecast (via DashCast)

marcosdiez:
In another step to make `catt` cast ALL the things, here is a simple PR to make it cast any website to Chromecast.

The heavy lifting is done by DashCast ( http://stestagg.github.io/dashcast/ ), which was easier than I thought to implement on `catt` because `pychromecast` already supports it.

I decided not to change the UI by falling back to `dastcast` every time `youtube-dl` can not handle a URL.

That means you just needs to type:

`catt cast https://www.github.com/` 

to see github on your chromecast!

`catt info` shows the title of the webpage that is being displayed.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -13,7 +13,10 @@ from .util import warning
 from .youtube import YouTubeController
 
 
-APP_INFO = [{"app_name": "youtube", "app_id": "233637DE", "supported_device_types": ["cast"]}]
+APP_INFO = [
+    {"app_name": "youtube", "app_id": "233637DE", "supported_device_types": ["cast"]},
+    {"app_name": "dashcast", "app_id": DASHCAST_APP_ID, "supported_device_types": ["cast"]},
+]
 DEFAULT_APP = {"app_name": "default", "app_id": "CC1AD845"}
 DASHCAST_APP = {"app_name": "dashcast", "app_id": DASHCAST_APP_ID}
 BACKDROP_APP_ID = "E8C28D3C"
@@ -83,10 +86,8 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
     cast.wait()
 
     if controller == "dashcast":
-        cc_info = (cast.device.manufacturer, cast.model_name)
         controller = DashCastController(cast, DASHCAST_APP["app_name"], DASHCAST_APP["app_id"], prep=prep)
-        stream = StreamInfo(video_url, model=cc_info, is_standard_website=True)
-        return (controller, stream)
+        return controller
 
     if video_url:
         cc_info = (cast.device.manufacturer, cast.model_name)
@@ -344,7 +345,7 @@ class CastController:
     def _prep_control(self):
         """Make shure chromecast is in an active state."""
 
-        if self._cast.app_id == DASHCAST_APP["app_id"]:
+        if self._cast.app_id == DASHCAST_APP_ID:
             return
         if self._cast.app_id == BACKDROP_APP_ID or not self._cast.app_id:
             raise CattCastError("Chromecast is inactive.")
@@ -463,10 +464,9 @@ class DashCastController(CastController):
     def __init__(self, cast, name, app_id, prep=None):
         self._controller = PyChromecastDashCastController()
         super(DashCastController, self).__init__(cast, name, app_id, prep=prep)
-        self.info_type = "url"
 
-    def play_media_url(self, video_url, **kwargs):
-        self._controller.load_url(video_url, force=True)
+    def load_url(self, url, **kwargs):
+        self._controller.load_url(url, force=True)
 
 
 class YoutubeCastController(CastController):
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -82,12 +82,15 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
         cache.set_data(cast.name, cast.host)
     cast.wait()
 
+    if controller == "dashcast":
+        cc_info = (cast.device.manufacturer, cast.model_name)
+        controller = DashCastController(cast, DASHCAST_APP["app_name"], DASHCAST_APP["app_id"], prep=prep)
+        stream = StreamInfo(video_url, model=cc_info, is_standard_website=True)
+        return (controller, stream)
+
     if video_url:
         cc_info = (cast.device.manufacturer, cast.model_name)
-        stream = StreamInfo(video_url, model=cc_info, host=cast.host)
-        if stream.is_standard_website:
-            controller = DashCastController(cast, DASHCAST_APP["app_name"], DASHCAST_APP["app_id"], prep=prep)
-            return (controller, stream)
+        stream = StreamInfo(video_url, model=cc_info)
 
     if controller:
         if controller == "default":
@@ -118,6 +121,7 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
         controller = YoutubeCastController(cast, app["app_name"], app["app_id"], prep=prep)
     else:
         controller = DefaultCastController(cast, app["app_name"], app["app_id"], prep=prep)
+
     return (controller, stream) if stream else controller
 
 
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -231,23 +231,20 @@ class CastState(CattStore):
 
 class CastStatusListener:
     def __init__(self, app_id, active_app_id):
-        self.app_id = app_id
         self.app_ready = threading.Event()
-        self.backdrop_ready = threading.Event()
-        if app_id == BACKDROP_APP_ID:
-            self.backdrop_ready.set()
+        self.set_app_id(app_id, active_app_id)
+
+    def set_app_id(self, app_id, active_app_id=None):
+        self.app_id = app_id
         if app_id == active_app_id:
             self.app_ready.set()
+        else:
+            self.app_ready.clear()
 
     def new_cast_status(self, status):
         if self._is_app_ready(status):
             self.app_ready.set()
-            self.backdrop_ready.clear()
         else:
-            if status.app_id == BACKDROP_APP_ID:
-                self.backdrop_ready.set()
-            else:
-                self.backdrop_ready.clear()
             self.app_ready.clear()
 
     def _is_app_ready(self, status):
@@ -483,8 +480,10 @@ class DashCastController(CastController):
     def _prep_app(self):
         """Make sure desired chromecast app is running."""
         if self._cast.app_id == DASHCAST_APP_ID:
+            self._cast_listener.set_app_id(BACKDROP_APP_ID, self._cast.app_id)
             self.kill()
-            self._cast_listener.backdrop_ready.wait()
+            self._cast_listener.app_ready.wait()
+            self._cast_listener.set_app_id(DASHCAST_APP_ID)
 
         self._cast.start_app(DASHCAST_APP_ID)
         self._cast_listener.app_ready.wait()
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -480,6 +480,9 @@ class DashCastController(CastController):
     def _prep_app(self):
         """Make sure desired chromecast app is running."""
         if self._cast.app_id == DASHCAST_APP_ID:
+            # After a URL is loaded, the loaded page assumes control of Chromecast
+            # and therefore ignore any future command we to DashCast.
+            # Albeit annoying, the solution is quite simple: we kill the dashcast app.
             self._cast_listener.set_app_id(BACKDROP_APP_ID, self._cast.app_id)
             self.kill()
             self._cast_listener.app_ready.wait()
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -477,7 +477,7 @@ class DashCastController(CastController):
         self._load_url(url)  # this will actually take us to the "Waiting for URL" screen
         while not self._is_waiting_for_url():
             time.sleep(1)
-        self._load_url(url)
+        self._load_url(url)  # this will finally load the URL
 
     def _load_url(self, url, **kwargs):
         self._controller.load_url(url, force=True)
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -121,7 +121,6 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
         controller = DashCastController(cast, app["app_name"], app["app_id"], prep=prep)
     else:
         controller = DefaultCastController(cast, app["app_name"], app["app_id"], prep=prep)
-
     return (controller, stream) if stream else controller
 
 
@@ -231,9 +230,9 @@ class CastState(CattStore):
 
 class CastStatusListener:
     def __init__(self, app_id, active_app_id):
+        self.app_id = app_id
         self.app_ready = threading.Event()
         self.backdrop_ready = threading.Event()
-        self.app_id = app_id
         self._clear_or_set_backdrop_event(active_app_id)
         if app_id == active_app_id:
             self.app_ready.set()
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -234,8 +234,6 @@ class CastStatusListener:
         self.app_ready = threading.Event()
         if app_id == active_app_id and app_id != DASHCAST_APP_ID:
             self.app_ready.set()
-        else:
-            self.app_ready.clear()
 
     def new_cast_status(self, status):
         if self._is_app_ready(status):
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -5,6 +5,8 @@ from pathlib import Path
 
 import pychromecast
 from click import ClickException, echo
+from pychromecast.controllers.dashcast import APP_DASHCAST as DASHCAST_APP_ID
+from pychromecast.controllers.dashcast import DashCastController as PyChromecastDashCastController
 
 from .stream_info import StreamInfo
 from .util import warning
@@ -13,6 +15,7 @@ from .youtube import YouTubeController
 
 APP_INFO = [{"app_name": "youtube", "app_id": "233637DE", "supported_device_types": ["cast"]}]
 DEFAULT_APP = {"app_name": "default", "app_id": "CC1AD845"}
+DASHCAST_APP = {"app_name": "dashcast", "app_id": DASHCAST_APP_ID}
 BACKDROP_APP_ID = "E8C28D3C"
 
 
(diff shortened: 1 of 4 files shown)


skorokithakis:
Thank you for the PR! I'm a bit apprehensive about the mode of failure where the URL mistakenly doesn't return a video and the site itself gets cast. Maybe this should be an explicit switch that would force casting a site?

@theychx, what do you think?

mdiez-modus:
I though a lot about that. Maybe have a completely different command, like `catt cast_url URL` 
I don't like heuristics either.
Since this is a delicate topic in which is very easy for the 3 of us to have at least 4 different opinions, I let you guys decide what the best interface should be.

I can easily implement it after a conclusion is achieved!

skorokithakis:
Yes, I think a dedicated command might be much better. I think this also fits with the spirit of catt, generally, and the code doesn't seem very complicated.

mdiez-modus:
my code is broken. please don't merge. I am adding the `cast_url`  command as well. please do not merge yet

theychx:
I'm not keen on the current approach. Making a separate command seems like the obvious solution, as this  would mean that you could just call ```setup_cast``` with ```controller="dashcast", prep="app"``` and not need to modify ```StreamInfo``` or ```setup_cast```.

skorokithakis:
Yes, I'm keen on that for the same reason, it would remove much of the `if is_website` code duplication that smells bad.

theychx:
Yeah, or just ```mirror```.

skorokithakis:
I think `mirror_site` is a bit more explicit as to what you're mirroring, but I have no strong preference either way.

marcosdiez:
Sorry @skorokithakis but that's completely different. What Chrome does is open the tab in itself and mirror what it has rendered to Chromecast.
What I plan to do is tell Chromecast to open a URL. There is no mirroring. The website is even accessed using Chromecast's own user agent.

skorokithakis:
Hmm, yes, good point. Do you want to make the changes first, and decide on the name later? Renaming the command is a one-line change anyway, and I think we have consensus that this is a useful feature to include.

marcosdiez:
The idea of making this a separate URL really made everything easier.

I did not like to have to change `StreamInfo` and the changes I did to `setup_cast` were minimal. 

I also made sure we can reload URLs, something that was previously not possible.

We now have only to decide how the command will be called.

I let to you guys decide. I am happy with the functionality!

Thanks for the code review

catt/controllers.py
```diff
@@ -447,6 +458,40 @@ def restore(self, data):
...
+            # because the new URL does not know it's on Chromecast and therefore does not answer
+            # so the only solution is to kill the DashCast app and load it again
+            self.kill()
+            while self._is_app_loaded():
+                time.sleep(1)
+
+        self._load_url(url)  # this will actually take us to the "Waiting for URL" screen
+        while not self._is_waiting_for_url():
+            time.sleep(1)
+        self._load_url(url)  # this will finally load the URL
```
theychx: I have found a way to make this work without going through these hoops:
Override ```_prep_app``` in your ```dashcast``` controller, so that it always kills the running app.
Add a ```backdrop_ready``` flag to ```cast_listener```.
In the overridden ```prep_app```, kill the running app and wait for the flag to be set.

catt/controllers.py
```diff
@@ -110,8 +116,11 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
```
theychx: We need to fail here, if the type of the chosen cast device is not in ```supported_device_types```.

catt/controllers.py
```diff
@@ -334,11 +343,13 @@ def _prep_app(self):
     def _prep_control(self):
         """Make shure chromecast is in an active state."""
 
+        if self._cast.app_id == DASHCAST_APP_ID:
+            return
```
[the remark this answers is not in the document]
mdiez-modus: This is not a hack. With this line, it's possible to type `catt info` and see the title of the current loaded webpage. Without it, we get a boring `Error: Nothing is currently playing.`.

catt/controllers.py
```diff
@@ -334,11 +343,13 @@ def _prep_app(self):
...
         if self._cast.app_id == BACKDROP_APP_ID or not self._cast.app_id:
             raise CattCastError("Chromecast is inactive.")
         self._cast.media_controller.block_until_active(1.0)
         if self._cast.media_controller.status.player_state in ["UNKNOWN", "IDLE"]:
-            raise CattCastError("Nothing is currently playing.")
+            raise CattCastError("Nothing is currently playing. [APP_ID={}]".format(self._cast.app_id))
```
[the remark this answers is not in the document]
mdiez-modus: That was useful to me, but I don't mind removing it.

catt/controllers.py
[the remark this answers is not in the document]
theychx: It also allows the user to use ```catt ffwd 60``` while a webpage is loaded, and then get the irrelevant  error message ```Error: Stream is not seekable.```
```info``` and ```status``` displays video/audio info (as descibed in the command helptexts).

File: catt/controllers.py (hunk 2 above)
Comment (mdiez-modus) [reply catt/controllers.py]: Instead of failing, I decided to just print a warning for two reasons:

- this is what `catt` currently do for other stuff
- warnings are more future proof than failures

File: catt/controllers.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (mdiez-modus) [reply catt/controllers.py]: ok... you convinced me!

**Comment by mdiez-modus** [comment]
@theychx thanks for the hints about using `_prep_app` and `CastStatusListener`

By using both of them properly, I was able to remove all every `time.sleep()` from `DashCastController`.

File: catt/controllers.py [hunk 5]
```diff
@@ -221,17 +231,29 @@ def get_data(self, name):
 
 class CastStatusListener:
     def __init__(self, app_id, active_app_id):
-        self.app_id = app_id
         self.app_ready = threading.Event()
+        self.set_app_id(app_id, active_app_id)
```
theychx: Please just add a ```backdrop_ready``` flag of the ```threading.Event``` kind, and then initiate it like ```app_ready```, and expand the logic in ```new_cast_status``` so the flag is set/cleared in the same manner as ```app_ready```.

catt/controllers.py
```diff
@@ -447,6 +467,31 @@ def restore(self, data):
...
+            # After a URL is loaded, the loaded page assumes control of Chromecast
+            # and therefore ignore any future command we to DashCast.
+            # Albeit annoying, the solution is quite simple: we kill the dashcast app.
+            self._cast_listener.set_app_id(BACKDROP_APP_ID, self._cast.app_id)
+            self.kill()
+            self._cast_listener.app_ready.wait()
+            self._cast_listener.set_app_id(DASHCAST_APP_ID)
+
+        self._cast.start_app(DASHCAST_APP_ID)
+        self._cast_listener.app_ready.wait()
```
theychx: I imagine ```_prep_app``` being as simple as this:

```
self.kill()
self._cast_listener.backdrop_ready.wait()
self._cast.start_app(self._cast_listener.app_id)
self._cast_listener.app_ready.wait()
```

marcosdiez:
@theychx I did all the changes you requested. `_prep_app` has one extra line than what you had in mind, because there is no need to kill the current app unless it's dashcast itself.

I believe it should be ready to be merged now!

catt/controllers.py
theychx: What I wanted was this:

```
class CastStatusListener:
    def __init__(self, app_id, active_app_id):
        self.app_id = app_id
        self.app_ready = threading.Event()
        self.backdrop_ready = threading.Event()
        if active_app_id == app_id:
            self.app_ready.set()
        elif active_app_id == BACKDROP_APP_ID:
            self.backdrop_ready.set()

    def new_cast_status(self, status):
        if status.app_id == self.app_id:
            self.app_ready.set()
            self.backdrop_ready.clear()
        elif status.app_id == BACKDROP_APP_ID:
            self.backdrop_ready.set()
            self.app_ready.clear()
        else:
            self.app_ready.clear()
            self.backdrop_ready.clear()
```

You might want to know that after the listener has been registered in the constructor of the ```CastController``` baseclass, ```pychromecast``` automatically calls the ```new_cast_status``` method of the listener object whenever there is a change.
Also, ```threading.Event``` flags are initially set to false.

marcosdiez:
I see. It all makes sense. By the way I was reading pychromecast's source code and I think there is a way to bypass backdrop completelly. I will test tomorrow after work

marcosdiez:
@theychx now it's G R E A T !
No more going to backdrop and going back to DashCast! Pages load a few seconds faster because of that.
Of course I did not like the internal API call I had to do. But that's what we have for today.

I will soon write a patch to pychromecast.

skorokithakis:
That's pretty fantastic. Hopefully your patch will stabilize the API so it doesn't change from under our feet.

marcosdiez:
My PR to `pychromecast` is here: https://github.com/balloob/pychromecast/pull/237
Now it's up to them to merge it!

catt/controllers.py
```diff
@@ -447,6 +461,25 @@ def restore(self, data):
...
+    def load_url(self, url, **kwargs):
+        self._controller.load_url(url, force=True)
+
+    def _prep_app(self):
+        """Make sure desired chromecast app is running."""
+        # we must force the launch of the DashCast app because it, by design,
+        # becomes unresponsive after a website is loaded
+        self._cast.socket_client.receiver_controller.launch_app(self._cast_listener.app_id, force_launch=True)
+        self._cast.start_app(self._cast_listener.app_id)
+        self._cast_listener.app_ready.wait()
```
theychx: If you clear the ```app_ready``` flag before force launching the app, it would not be necessary to modify ```CastStatusListener``` or use the class variable. Also the ```self._cast.start_app``` call is not needed anymore.

marcosdiez: Wouldn't not modifying `CastStatusListener` make things rely on a race condition ?
The modified version of `CastStatusListener` guarantees that the app is only ready when it *should* be.

The DashCast app has an "Application Is Starting" state in which it can not yet receive URLs: https://github.com/stestagg/dashcast/blob/master/receiver.html#L163

This is one of the reasons, actually, that I liked that `catt info` showed the state of the DashCast app.

theychx: > Wouldn't not modifying CastStatusListener make things rely on a race condition ?

Hmm, I guess thats true, although it's extremely unlikely that this would ever be a real problem. I've tested my own suggestion and could not get it to fail.
I'm keen on the code needed to circumvent the odd behaviour of dashcast, being contained within the overridden ```_prep_app```. And I am honestly not keen on the class variable and the use of it in ```CastStatusListener```.

@skorokithakis I'm not completely sure I'm in the right here, any input?

marcosdiez: I did not like the change either on `CastStatusListener` but either we leave things as is or we patch `DashCast` to behave as expected. Race conditions are must always be avoided because debugging them is a *real* pain. My vote is to do nothing and push the merge button.

skorokithakis: I'll have to agree with @marcosdiez here, if eliminating a race condition means that the code will be a bit uglier, I can live with that, since it will be really hard to debug something like "Sometimes catt doesn't work" or "catt doesn't work on my new Chromecast".

I would, however, like to add comments here and on `CastStatusListener` explaining exactly which race condition this guards against, and how it does that. I can definitely see us coming back in a year, not remembering this discussion and saying "Oh we don't actually need this, we can just clear the `app_ready` flag" and fall into the race condition trap.

theychx: The (very) remote possibility of a race condition would only occur when using dashcast, but fair enough.
I still don't understand why the current logic in ```CastStatusListener``` is needed, when we could just do this in the constructor:

```
if app_id == active_app_id and active_app_id != DASHCAST_APP_ID:
    self.app_ready.set()
```
This would have the same effect as my previous suggestion without introducing the risk of a race condition. Am I missing something?

skorokithakis: Hmm, it's been a while since I visited this part of the code and can't remember the flow here. Can either of you refresh my memory on why the race condition would happen?

skorokithakis: I see, thank you. I would still be in favor of preventing the race condition, but your proposed solution seems good to me, as I'm not a fan of modifying `CastStatusListener` either. @marcosdiez, do you see any reason why @theychx's suggestion wouldn't work?

marcosdiez: The whole problem arises because the DashCast app is odd.

It has two idiosyncrasies:

a) once loaded it, by design becomes unresponsive (because it redirects to the casted webpage which has no idea it's being rendered in Chromecast) (this has been solved previously by killing it and now by force loading the app every time)

b) when you load the chromecast app, it's initial state is `Application is Starting` https://github.com/stestagg/dashcast/blob/master/receiver.html#L163 

Only eventually the status becomes `Application Ready` https://github.com/stestagg/dashcast/blob/master/receiver.html#L143

The changes I did in `CastStatusListener` (which in the end of the day are not that ugly), explicitly set the state to ready only when it's really ready (by waiting for the `Application Ready` status).

And since I live in Brazil, which is the other side of earth, the internet has much more latency than in Europe and the US because we are very far and light is not that fast.

Australians have it worse. As soon as `catt` becomes a worldwide hit they will fill most of the bugs if we leave the race condition!

I am completely against leaving a race condition there. It will make casting a url something that will randomly fail, more often if there is more latency and after we forget all this, we will never ever fix it because, as with every race condition, even reproducing is a challenge. We know the problem, we have the solution ready and we are willing not to use it. Sorry to say that, but this is negligence. 

Thinking about it, the discussion here is the most technical solution versus the least intrusive solution. The former is always my choice.

`DashCast` needs special treatment and that's clearly defined in `CastStatusListener`. We can't hide away from that fact. `CastStatusListener` is the right place to deal with it.

Also, any coder in the future will see that. I can, of course, improve the comments, but even if we leave the code as is, I don't believe anybody would ever delete any of the code from `CastStatusListener` by accident in the future, for cleanup's sake. (different from a flag that arbitrarily and unintuitively set's the ready state to false)

Guys, sorry for not agreeing. This is just my opinion and I do respect if you decide otherwise. Please don't get offended by me not agreeing.

skorokithakis: Hey man, nobody is offended, we're just discussing code. Let me read the rest of your comment and reply in more detail.

skorokithakis: @marcosdiez What is the status of other applications when they become ready?

skorokithakis: So here are my thoughts:

1. I'm really averse to having the race condition, I think we're all in the same page that we want to avoid that, even if the code gets a bit uglier.
1. If the other applications have the same status string as DashCast, maybe we can generalize and not hardcode it for just DashCast.
2. Since the `app_ready` string is a DashCast-specific hack, I'm not crazy about it being called `app_ready` (collides with another variable and the person reading the code gets confused). I would rather just have the code in `_is_app_ready` be `return status.status_text == "Application ready"` with a big comment saying what you said above about the race condition.
3. I'm still not sure why @theychx's suggestion wouldn't work. Can you elaborate a bit, @marcosdiez?

Other than that, I think this is very close to merging, great work!

theychx: @marcosdiez I think you missed my new proposal, which do not introduce the risk of a race condition. Please read that, and explain to us why that would not suffice. I have not been able to get this new suggestion to fail either.
Also, I just tested the signals coming from pychromecast to the listener, and it looks as though the app id updates don't fire before the app is truly ready. But I urge you to try and make my second suggestion fail, and if you succeed I will of course change my mind.

PS: I'm not offended. :smiley:

skorokithakis: @theychx To clarify, you mean we'd add that check to the `__init__` of `DashCastController`, right?

theychx: No, __init__ of ```CastStatusListener```.
@marcosdiez is right about the app id update firing while the status text is still "Application is starting". I'm just not a fan of remedying that, unless it becomes a real problem for users. Like I said, my second suggestion seems perfectly reliable.

theychx: @skorokithakis 
> If the other applications have the same status string as DashCast, maybe we can generalize and not hardcode it for just DashCast.

No other app I have tested has this behaviour.

skorokithakis: Also, it looks like your suggestion sets the `app_ready` flag if the app is not `DashCast`, but don't we need a way to tell that DashCast is ready? Sorry to be such a pain, I just don't understand the flow here exactly.

theychx: Yep, but he also implemented the extra status text check, and it is the necessity of this I am contesting.

skorokithakis: @theychx Can we discuss somewhere more real-time? I've created a Gitter room and invited you there (you can just log in with Github).

skorokithakis: @marcosdiez You're welcome too, of course.

skorokithakis: Hey @marcosdiez, @theychx and I discussed this on Gitter, and we agree that eliminating the race condition makes the most sense. However, could you change the code in the check to this?:

```python
def _is_app_ready(self, status):
    if status.app_id != self.app_id:
        return False
    if self.app_id == DASHCAST_APP_ID:
        # Big-ass comment about why this is here here.
        return status.status_text == "Application ready"
    return True
```

That way, we get rid of the confusingly-named class variable that's only used in one place, and we have a comment on the special case explaining exactly why it happens.

Also, could you add a comment on the check in `__init__` on the `CastStatusListener` about why the DashCast check is necessary? I think this is a good compromise between clearly documenting the special-case code and eliminating any potential race conditions.

marcosdiez:
Hi guys, 
sorry for not participating in the Gitter discussion. I was busy with my real job :/

@theychx you were right, `self._cast_listener.app_ready.wait()` was redundant and therefore was removed

@skorokithakis Yet on the race condition topic, I did check on the `YouTube` and `Default Media Player` app and they both already start in a `ready` status.

I did describe the race condition in the code as best as I could. Feel free to request further changes or modify yourself. 

And by the way, the triple comparison was a nice trick that I was not aware of. Thanks for teaching me that one!

skorokithakis:
Thank you, @marcosdiez! The comment looks good to me. However, you seem to have removed `start_app()` instead of `wait()`? Is that a mistake?

theychx:
It's cool. He just quoted the wrong line, the code is correct.

skorokithakis:
Ah alright, great. Are you good to merge this, @theychx?

theychx:
Yep

skorokithakis:
Fantastic! There are a few conflicts, it seems, which is a bit odd since we didn't do much to master. Is either of you at a computer and able to resolve those?

theychx:
It was just the gitter badge.

skorokithakis:
Thanks, I realized there was a button but it didn't show on mobile.

skorokithakis:
Merged, thank you both!

PR #103, opened 2018-05-25
[WIP] Prepare for 0.70

theychx:
Howdy
We forgot to make a final decision, with regards to the command name of the new dashcast functionality.
```cast_site``` or ```display_site``` are decent ideas.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,6 +43,9 @@ natively)::
 
     catt cast ./myvideo.mp4
 
+You can also control your Chromecast through ``catt`` commands, for example with
+``catt pause``. Try running ``catt --help`` to see the full list of commands.
+
 If you have subtitles and the name is similar to the name of the local file, ``catt`` will add them automatically.
 You can, of course, specify any other subtitle if you want. Although Chromecast only supports WEBVTT,
 TTML and Line 21 subtitles, ``catt`` conveniently converts SRTs to WEBVTT for you on the fly. Here is how to use it::
@@ -55,9 +58,6 @@ TTML and Line 21 subtitles, ``catt`` conveniently converts SRTs to WEBVTT for yo
 
 Please note that the Chromecast has a slow CPU but a reasonably recent version of Google Chrome. The display
 resolution is 1280x720.
-
-You can also control your Chromecast through ``catt`` commands, for example with
-``catt pause``. Try running ``catt --help`` to see the full list of commands.
 
 Configuration file
 """"""""""""""""""
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ natively)::
 
 If you have subtitles and the name is similar to the name of the local file, ``catt`` will add them automatically.
 You can, of course, specify any other subtitle if you want. Although Chromecast only supports WEBVTT,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by skorokithakis** [comment]
I like `cast_site`, I don't think it's so easy to conflate with `cast`, since it's so much longer.

**Comment by marcosdiez** [comment]
If I had to choose between the two, I would pick `cast_site`, but I am happy with either.

I actually thought you two were happy with `cast_url`.

Since we are discussing features, I thougt about making an interactive screen to play/pause/forward the video. Ehat do you guys think?

**Comment by skorokithakis** [comment]
`cast_site` and `cast_url` are fine with me, I'm just worried that the latter is a bit ambiguous, because casting a URL is also exactly what the `cast` command does, so it's not as clear. URLs can be sites, but they can be other things too (like videos).

What sort of screen do you mean? Like a persistent cli interface?

**Comment by marcosdiez** [comment]
I was thinking about somethink like this: https://sparkylinux.org/images/3/cli/sparky3-moc.png

Initially not as fancy.

So something that I could easily play/pause a video just with the space bar and use the arrows to advance /skip 10 seconds / 1 minute / 10 minutes / the track.

In other words, some interactive screen that allowes me to manipulate the current playing video without having to type 1000 keystokes for every command

File: catt/cli.py [hunk 1]
```diff
@@ -261,6 +249,15 @@ def cast(settings, video_url, subtitle, force_default, random_play, no_subs):
             cst.play_media_id(stream.video_id)
 
 
+@cli.command(short_help="Cast any webpage to a Chromecast.")
```
marcosdiez: Please`webpage` to `website`

skorokithakis:
I would love a CLI like that (like pianobar does), because I've needed it many times. I'm wondering whether it fits the purpose of `catt`, since `catt` is meant to cast things (although I guess we already offer control and this would be a simpler interface).

There's also the question of having things in the repo vs making the code more modular so utilities could be written on top of catt. In this case, I don't think the latter makes sense, since one could just use `pychromecast` directly and not need any of the functionality in `catt`. I would be in favor of a persistent interface that controls playback, I think.

@theychx, what do you think?

theychx:
I can definitely see this being useful to some users. I would very much prefer an approach that looks something like this:

The interactive stuff is in a seperate module.
The user would be activating an interactive session by doing:
```catt cast -i <url>```
At the end of the main clause in ```cli.cast``` we would simply do:
```
if interactive:
    do_interactive_stuff()
```
And for local file casting:
```
if not interactive:
    while thr.is_alive():
        time.sleep(1)
```
and then resolve any remaining issues in the interactive module.

If the implementation of this requires any significant changes to existing code (especially outside of ```cli.py```), then my enthusiasm for this will drop quite a lot.

skorokithakis:
I agree with this, I'd just like to add that we should have an extra command to launch just the interactive module if the user wants to, in case they forgot/changed their mind/didn't use `catt` to cast stuff.

I think it would be pretty doable to just do this without touching any other code, no?

theychx:
The playlist part of the main clause of ```cli.cast``` would need to change a bit (it looks weird anyway).
After ```0.7.0``` is released I plan to some spring cleaning / refactoring, so I'll addresss this too.

By the way, this PR is ready to merge, and should also be ready to release.

PR #104, opened 2018-05-26
Refactor playlist capability handling slightly

theychx:
..so we can put additional code after the main clause in cli.cast.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -16,7 +16,6 @@ from .controllers import (
     CastState,
     get_chromecast,
     get_chromecasts,
-    PlaybackError,
     setup_cast,
     StateFileError
 )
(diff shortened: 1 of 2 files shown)


catt/cli.py
```diff
@@ -218,25 +217,22 @@ def cast(settings, video_url, subtitle, force_default, random_play, no_subs):
...
+            cst.play_playlist(stream.playlist_all_ids)
         else:
-            try:
-                if not stream.playlist_all_ids:
-                    raise ValueError
-                cst.play_playlist(stream.playlist_all_ids)
-                return
-            except (PlaybackError, ValueError):
+            if random_play:
+                stream.set_playlist_entry(random.randrange(0, stream.playlist_length))
```
skorokithakis: Do you want to split the call to `set_playlist_entry()` here outside the `if` so it's not repeated?

theychx: I'm not sure I understand what you mean.

skorokithakis: Just:

```python
if random_play:
    entry = random.randrange(0, stream.playlist_length)
else:
    warning("Playlist playback not possible, playing first video.")
    entry = 0
stream.set_playlist_entry(entry)
```

theychx: Ah, ok. Hang on.

theychx: I'm watching formula 1, so I'm not entirely focused. :smile:

skorokithakis: Haha, understandable :P

PR #107, opened 2018-05-31
fix: Make dashcast work with audio devices

theychx:
This might sound like a silly idea, but it could be useful as a fallback for sites with autoplaying content that is not supported by ```youtube-dl```.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -14,7 +14,7 @@ from .youtube import YouTubeController
 
 APP_INFO = [
     {"app_name": "youtube", "app_id": "233637DE", "supported_device_types": ["cast"]},
-    {"app_name": "dashcast", "app_id": DASHCAST_APP_ID, "supported_device_types": ["cast"]},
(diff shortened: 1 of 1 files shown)


theychx:
I don't understand the CI failure. The hook ran fine when commiting.

skorokithakis:
Sorry, that one was on me. I forgot to include it in the dev requirements.

skorokithakis:
Hm, this is interesting. Have you tried to see if it works? I'm currently having issues discovering my Mi box at all.

theychx:
Yep, works with soundcloud.

theychx:
CI fails now for py <3.6 because of ```black```.

skorokithakis:
Yep, that's super inconvenient...

theychx:
How did you resolve this?

skorokithakis:
I just force-pushed, I'll resolve later as I'm very busy now :(

theychx:
I have some PR's coming up, most of them small. Is there any specific time you'd prefer to deal with this? I can also roll some of the changes into a couple of PR's, if you prefer.

skorokithakis:
Nah just send them all in and I'll get to them as I can.

PR #108, opened 2018-05-31
Make cast_site accept stdin as argument

theychx:
PS: ```flake8-import-order``` and ```isort``` do not agree on import order.


skorokithakis:
flake8-import-order should be removed, oops.

skorokithakis:
Oh it's just making `cast_site` more consistent. Got it, thanks.

PR #110, opened 2018-05-31
Add missing dots to messages

theychx:
..........


theychx:
No, I did that. :smiley:

skorokithakis:
Oh cool, if black is fine with it, so am I!

PR #111, opened 2018-06-01
fix: Make save restore more robust

theychx:
A few improvements:
+ We now fail with the correct error message, if the user tries ```catt restore```, without any config dir save file having been created.
+ I've made an attempt at making the create dir/file logic less flakey.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -362,7 +362,7 @@ def save(settings, path):
     if path and path.is_file():
         click.confirm("File already exists. Overwrite?", abort=True)
     click.echo("Saving...")
-    state = CastState(path or STATE_PATH, create_dir=True if not path else False)
+    state = CastState(path or STATE_PATH, mode="arbi" if path else "conf")
     state.set_data(cst.cc_name, {"controller": cst.name, "data": cst.media_info})
 
 
@@ -370,6 +370,8 @@ def save(settings, path):
 @click.argument("path", type=click.Path(exists=True), callback=process_path, required=False)
 @click.pass_obj
 def restore(settings, path):
+    if not path and not STATE_PATH.is_file():
+        raise CattCliError("Save file in config dir has not been created.")
     cst = setup_cast(settings["device"])
     state = CastState(path or STATE_PATH)
     try:
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -11,7 +11,7 @@ from threading import Thread
 import click
 import requests
 
-from .controllers import Cache, CastState, StateFileError, get_chromecast, get_chromecasts, setup_cast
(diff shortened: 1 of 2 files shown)


catt/cli.py
```diff
@@ -362,14 +362,16 @@ def save(settings, path):
     if path and path.is_file():
         click.confirm("File already exists. Overwrite?", abort=True)
     click.echo("Saving...")
-    state = CastState(path or STATE_PATH, create_dir=True if not path else False)
+    state = CastState(path or STATE_PATH, mode="arbi" if path else "conf")
```
skorokithakis: How this is going to be parsed is pretty confusing, it needs parentheses to clarify.

catt/controllers.py
```diff
@@ -205,12 +205,18 @@ def get_data(self, name):
 
 
 class CastState(CattStore):
-    def __init__(self, state_path, create_dir=False):
+    def __init__(self, state_path, mode="read"):
```
skorokithakis: Hmm, why is `mode` here a string rather than an `enum` or something similar? A string seems pretty error-prone.

theychx: Something like?:
```
class Mode(enum.Enum):
    READ = auto()
    CONF = auto()
    ARBI = auto()
```
Forgive my ignorance, but why would the string be error-prone?

skorokithakis: The way you've implemented it, it'll raise an error, which is good, but I've seen this be implemented as "if x: elif y: else:", and then people misspelled "conf" and boom, the mode was "read". It's also less semantic (`def myfunc(mode: str)` is much less readable than `def myfunc(mode: StateMode)`) and the type checker can't check it as easily.

theychx: Thank you for the explanation. Did you like my enum example?

skorokithakis: Sorry, yes, that's perfect. I would name it something more descriptive than `Mode` ("what kind of mode?"), but other than that it's great.

catt/cli.py
theychx: Around both parts, or just the latter?

catt/controllers.py
skorokithakis: I really wish they had something like `MyEnum = enum_generator(["foo", "bar", "baz"])` instead of the heavier class-based approach, but oh well.

catt/cli.py
skorokithakis: Oh, I misread this due to the complexity. I would just split it up:

```
if path:
    state = CastState(path, mode.ARBI)
else:
    state = CastState(STATE_PATH, mode.CONF)
```

theychx:
Please merge.

skorokithakis:
Ah yes, forgot about the failure :/

PR #115, opened 2018-06-02
Make use of enums in CastState even safer

theychx:
It occurs to me that we would have to do something like this, for this to be safe. Or just check for all possible attributes of the enum and then do ```else fail```, like I did before. Is there any convention/best practice when doing this kind of stuff in python?

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -213,6 +213,8 @@ class StateMode(Enum):
 
 class CastState(CattStore):
     def __init__(self, state_path, mode):
+        if not isinstance(mode, StateMode):
+            raise ValueError("invalid mode")
         super(CastState, self).__init__(state_path)
         if mode == StateMode.CONF:
             self._create_store_dir()
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -212,9 +212,7 @@ class StateMode(Enum):
 
 
 class CastState(CattStore):
-    def __init__(self, state_path, mode):
-        if not isinstance(mode, StateMode):
-            raise ValueError("invalid mode")
+    def __init__(self, state_path: Path, mode: StateMode):
         super(CastState, self).__init__(state_path)
         if mode == StateMode.CONF:
             self._create_store_dir()
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -212,7 +212,7 @@ class StateMode(Enum):
 
 
 class CastState(CattStore):
-    def __init__(self, state_path: Path, mode: StateMode):
+    def __init__(self, state_path: Path, mode: StateMode) -> None:
         super(CastState, self).__init__(state_path)
         if mode == StateMode.CONF:
             self._create_store_dir()


theychx:
So we need the extra dependency to check type, right? Then what is the purpose of the above syntax?
(and what would be your preference between mypy and typeguard?)

skorokithakis:
It's a dev-time dependency, like flake8. The purpose of the syntax is that now you can both check it with a tool, *and* know by just looking at it that `state_path` is a `pathlib.Path` object (and `mode` is a `StateMode`), so it's pretty much great documentation.

It's not like the code will work fine in development and will suddenly start passing integers for the `mode` in production anyway :stuck_out_tongue:

skorokithakis:
As for the preference between `mypy` and `typeguard`, I usually use both in standalone apps (where I control the distribution) and only use `mypy` in libraries, to avoid the extra runtime dependency.

skorokithakis:
I've added `mypy` checks to the repo and helped initial checks pass. You can merge master into this branch and add types to a few things that you find. You can also play around locally with `mypy -s catt/` to see how things break with invalid types, how PyCharm (if you use it) becomes type-aware and suggests the right thing to pass into the function, etc.

theychx:
Amazing, thanks! (and I use VS Code)

skorokithakis:
Ah, yep. `mypy` already found a few bugs on 3.4 :/

skorokithakis:
What are we doing with this PR? Do you want to turn it into a more general "add a few types here and there" PR, or just add the type in this single call?

theychx:
We might as well do the former.

skorokithakis:
Yeah, agreed, at least in some places where it's obvious what we pass.

theychx:
:confused: 
From Travis (that succeeded on 3.4, but failed on 3.5 and 3.6):
```
catt/controllers.py:215: error: The return type of "__init__" must be None
```

skorokithakis:
Oh, just add `-> None:` at the end, that should fix it.

skorokithakis:
`def __init__(self, state_path: Path, mode: StateMode) -> None:`

theychx:
But isn't that the ```typing``` syntax that was introduced in 3.5?

skorokithakis:
No, the `typing` module was introduced in 3.5, function annotations have always been there in 3. As long as you don't use any of the types in the `typing` module (which in my opinion is alone a good reason to deprecate 3.4 support), it should run fine on 3.4.

theychx:
Ok.
> which in my opinion is alone a good reason to deprecate 3.4 support
So why don't we deprecate 3.4 support. :smiley:

skorokithakis:
I don't love deprecating things for less-than-good reason. We aren't currently actually *using* `typing`, we can deprecate if we ever need to :P

theychx:
Well, it was worth a try, wasn't it. :smile:

skorokithakis:
I'm just waiting for the unused `typing` import now that you'll use to justify deprecating stuff :P

theychx:
With the ```mypy``` issues adressed in https://github.com/skorokithakis/catt/pull/120/commits/1f88a6f14e1d2f82ab3c4cc8e961ab8d40b1355e, I'm a bit uncertain on how to proceed with this PR. Would you mind merging this now, and then revisit later?

skorokithakis:
Sure, I can merge, but won't that make CI fail? Do we need to upgrade pre-commit or something?

theychx:
I'm not shure. The two times I merged master back into this branch, things were fine.

skorokithakis:
Let's see what happens!

PR #116, opened 2018-06-04
Fix the Travis build

skorokithakis:


theychx:
Why did Travis fail with 3.4?

skorokithakis:
I have no idea, mypy thinks that a `Path` object isn't a valid argument to `Path()`...

PR #120, opened 2018-07-16
Group support

### LINKED ISSUE #119: Support chromecast audio groups
[`get_chromecasts` in controllers.py](https://github.com/skorokithakis/catt/blob/master/catt/controllers.py#L32) ensures that all devices can be dialed by creating a new `pychromecast.Chromecast` using the device's `host` parameter. Unfortunately, for audio groups, the `host` param is mostly meaningless and chosen arbitrarily from the devices that participate in the group. This is easily visible from the output of `catt scan`:
```
> catt scan
Scanning Chromecasts...
192.168.1.75 - Bedroom - Google Inc. Google Home
192.168.1.77 - Bedroom TV - Google Inc. Eureka Dongle
192.168.1.70 - Kitchen - Google Inc. Google Home Mini
192.168.1.76 - Living Room - Google Inc. Chromecast Audio
192.168.1.70 - Kitchen - Google Inc. Google Home Mini
```
In this output it's the Kitchen device that's present twice, but this varies across runs as different elements of the group are selected.

Since all device lookups go through `get_chromecasts`, this means that any command directed at an audio group instead operates on a single arbitrary member of the group.

It _appears_ that removing the list comprehension in the final statement of `get_chromecasts` (that is, just return `devices` instead of rebuilding each one with an address) fixes this and lets catt play to chromecast audio groups, but since I don't know why that functionality was added in the first place and was too lazy to get a dev env properly set up, I'll submit a feature request instead of a PR.

theychx:
We are now assuming that 1st/2nd gen. chromecasts are the only kinds of devices that report different names via DIAL/zeroconf.

fixes #119

Please test, as I do not have access to multiple audio devices myself.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -28,8 +28,7 @@ def get_chromecasts(fail=True):
         raise CattCastError("No devices found.")
 
     devices.sort(key=lambda cc: cc.name)
-    # We need to ensure that all Chromecast objects contain DIAL info.
-    return [pychromecast.Chromecast(cc.host) for cc in devices]
+    return devices
 
 
 def get_chromecast(device_name):
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -21,10 +21,10 @@ DEFAULT_APP = {"app_name": "default", "app_id": "CC1AD845"}
 BACKDROP_APP_ID = "E8C28D3C"
 
 
-def get_chromecasts():
+def get_chromecasts(fail=True):
     devices = pychromecast.get_chromecasts()
 
-    if not devices:
+    if fail and not devices:
         raise CattCastError("No devices found.")
 
     devices.sort(key=lambda cc: cc.name)
@@ -189,7 +189,7 @@ class Cache(CattStore):
         self._create_store_dir()
 
         if not self.store_path.is_file():
-            devices = pychromecast.get_chromecasts()
+            devices = get_chromecasts(fail=False)
             self._write_store({d.name: d.host for d in devices})
 
     def get_data(self, name):
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -21,6 +21,7 @@ APP_INFO = [
 DEFAULT_APP = {"app_name": "default", "app_id": "CC1AD845"}
 BACKDROP_APP_ID = "E8C28D3C"
 DEVICES_WITH_TWO_MODEL_NAMES = {"Eureka Dongle": "Chromecast"}
+DEFAULT_PORT = 8009
 
 
 def get_chromecasts(fail=True):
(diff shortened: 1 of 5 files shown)


skorokithakis:
Thanks, @theychx! @saulrh, can you give this a spin?

theychx:
@skorokithakis 
We would have to store the group port in the cache, for this to work properly (there is no default group port). I don't know how you want to handle this, with regards to upgrades.

skorokithakis:
Hmm, why would upgrades be an issue? We'd just need to make a note somewhere near the cache to know what kind of information we store, so we can migrate it (or just delete it) between versions.

It might not be such a bad idea to key the cache off the app version, if we ever introduce a backwards-incompatible schema change.

theychx:
OK, so we implement a check of the cache format and delete if it's wrong?

> It might not be such a bad idea to key the cache off the app version

Please explain.

skorokithakis:
Add the app's version to the keys of the cache (or the cache dir), so if the user upgrades the app to `0.9.3` or whatever, the cache dir `catt-0.9.3` won't be there.

theychx:
Ah ok, thanks.

saulrh:
It works! Thanks!

theychx:
@skorokithakis Did you forget about this?

skorokithakis:
No, it's still WIP so I thought you wanted to add something.

theychx:
Ah, didn't think about that. It's ready for review.

catt/controllers.py
```diff
@@ -184,25 +189,39 @@ def clear(self):
 
 class Cache(CattStore):
     def __init__(self):
-        cache_path = Path(tempfile.gettempdir(), "catt_cache", "chromecast_hosts")
+        cache_path = Path(tempfile.gettempdir(), "catt_%s_cache" % CATT_VERSION, "chromecast_hosts")
```
skorokithakis: Can you sha1 the version here and get the first 8 chars or so? It's not very likely, but I'd like to guard against a version containing slashes like `0.4.3-pre1/amd` or whatever.

catt/controllers.py
```diff
@@ -184,25 +189,39 @@ def clear(self):
...
+            return (None, None)
+        if name:
+            fetched = data.get(name)
+        else:
+            # When the user does not specify a device, we need to make an attempt
+            # to consistently return the same IP, thus the alphabetical sorting.
+            fetched = data[min(data, key=str)]
+        return (fetched["ip"], fetched.get("group_port")) if fetched else (None, None)
+
+    def set_data(self, name, ip, port):
```
skorokithakis: Since we're 3-only now, can you add some types here? It helps with mypy. I'll add pre-commit so this is checked on CI as well.

skorokithakis: Oh we already added `pre-commit`, I'll see about adding mypy to it. The problem with having tens of projects is that you forget what you did where :/

catt/controllers.py
theychx: Yep

catt/controllers.py
theychx: Ok

theychx: I've had to comment out the baseclass ```set_data```  method in order to avoid getting this ```mypy``` error:
```
catt/controllers.py:221: error: Signature of "set_data" incompatible with supertype "CattStore"
```
(adding ```-> None``` does not help)
I realize that this might be bad OOP design anyway, so I'd appreciate any suggestions.

catt/controllers.py
theychx: Done

catt/controllers.py
```diff
@@ -221,7 +240,7 @@ def __init__(self, state_path, mode):
         elif mode == StateMode.ARBI:
             self._write_store({})
 
-    def get_data(self, name):
+    def get_data(self, name: str) -> str:
```
theychx: I didn't intend for this to be included in the commit.

skorokithakis: Which do you mean?

theychx: Just the change to this line.

catt/controllers.py
skorokithakis: I think this is a mypy bug. I added a type ignore on those lines for now.

theychx:
The ```typing``` module py3.5+.

skorokithakis:
Arghhhh

PR #122, opened 2018-07-21
YoutubeDL options implemented

xybydy:
Some of the websites that youtube-dl supports requires login. Read the catt's code but haven't found any solution or workaround for login, so I implemented it.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -41,6 +41,19 @@ class CattTimeParamType(click.ParamType):
 CATT_TIME = CattTimeParamType()
 
 
+class YtdlOptParamType(click.ParamType):
+    def convert(self, value, param, ctx):
+        if "=" not in value:
+            self.fail("%s is not a valid key value pair." % value, param, ctx)
+
+        ykey, yval = value.split("=", 1)
+        yval = {"true": True, "false": False}.get(yval.lower(), yval)
+        return (ykey, yval)
+
+
+YTDL_OPT = YtdlOptParamType()
+
+
 def human_time(seconds: int):
     return time.strftime("%H:%M:%S", time.gmtime(seconds))
 
@@ -176,12 +189,15 @@ def process_subtitle(ctx, param, value):
 @click.option(
     "--no-subs", is_flag=True, default=False, help="Don't try to load subtitles automatically from the local folder."
 )
+@click.option("-y", "--ytdl-option", type=YTDL_OPT, multiple=True, metavar="OPT", help="Youtube-dl option.")
 @click.pass_obj
-def cast(settings, video_url, subtitle, force_default, random_play, no_subs):
+def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_option):
     controller = "default" if force_default else None
     subtitle_url = None
     playlist_playback = False
-    cst, stream = setup_cast(settings["device"], video_url=video_url, prep="app", controller=controller)
+    cst, stream = setup_cast(
+        settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_option=ytdl_option
+    )
 
     if stream.is_local_file:
         click.echo("Casting local file %s..." % video_url)
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -191,19 +191,19 @@ def process_subtitle(ctx, param, value):
 )
 @click.option(
     "-y",
-    "--ytdl-options",
+    "--ytdl-option",
     type=YTDL_OPT,
     multiple=True,
     metavar="OPT",
-    help="Youtube-dl option. Should be passed as `-y key=value` format",
+    help="YouTube-DL option. Should be passed as `-y option=value`, and can be specified multiple times",
 )
 @click.pass_obj
-def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_options):
+def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_option):
     controller = "default" if force_default else None
     subtitle_url = None
     playlist_playback = False
     cst, stream = setup_cast(
-        settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_options=ytdl_options
+        settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_options=ytdl_option
     )
 
     if stream.is_local_file:
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -189,14 +189,21 @@ def process_subtitle(ctx, param, value):
 @click.option(
     "--no-subs", is_flag=True, default=False, help="Don't try to load subtitles automatically from the local folder."
 )
-@click.option("-y", "--ytdl-option", type=YTDL_OPT, multiple=True, metavar="OPT", help="Youtube-dl option.")
+@click.option(
+    "-y",
+    "--ytdl-options",
+    type=YTDL_OPT,
+    multiple=True,
+    metavar="OPT",
+    help="Youtube-dl option. Should be passed as `-y key=value` format",
+)
 @click.pass_obj
-def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_option):
+def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_options):
     controller = "default" if force_default else None
     subtitle_url = None
     playlist_playback = False
     cst, stream = setup_cast(
-        settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_option=ytdl_option
+        settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_options=ytdl_options
     )
 
     if stream.is_local_file:
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -44,10 +44,10 @@ CATT_TIME = CattTimeParamType()
 class YtdlOptParamType(click.ParamType):
     def convert(self, value, param, ctx):
         if "=" not in value:
-            self.fail("%s is not a valid key value pair." % value, param, ctx)
+            self.fail("%s is not a valid key/value pair." % value, param, ctx)
 
         ykey, yval = value.split("=", 1)
-        yval = {"true": True, "false": False}.get(yval.lower(), yval)
+        yval = {"true": True, "false": False}.get(yval.lower().strip(), yval)
         return (ykey, yval)
 
 
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -47,7 +47,7 @@ def get_chromecast(device_name):
         return devices[0]
 
 
-def setup_cast(device_name, video_url=None, prep=None, controller=None):
+def setup_cast(device_name, video_url=None, prep=None, controller=None, ytdl_option=None):
     """
     Prepares selected chromecast and/or media file.
 
@@ -70,6 +70,8 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
                        controller is bypassed, and the one specified here is
                        returned instead.
     :type controller: str
+    :param ytdl_option: YoutubeDL options
+    :type ytdl_option: tuple
     :returns: controllers.DefaultCastController or controllers.YoutubeCastController,
               and stream_info.StreamInfo if video_url is supplied.
     """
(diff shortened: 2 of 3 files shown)


skorokithakis:
Thank you! I'm a bit wary of proxying settings to YouTube-dl (if we're going to proxy some, might as well proxy them all), I will review this soon and return with feedback.

skorokithakis:
Alright, I've reviewed this and I would prefer it if we could pass YouTube-DL options in some other way. However, I'm not sure exactly how we could do this, since we're using the API and not calling the binary directly. I was thinking of something like `--ytdl-options="foo: bar, baz: ban"`, but that's clearly problematic when the options themselves contain colons or commas. Still, I think that adding an option to catt for every option in YTDL is going to grow unwieldy.

I'd appreciate suggestions from you guys, @xybydy and @theychx..

theychx:
We could do this:
```
@click.option("-y", "--ytdl-option", type=(str, str), multiple=True, help="Youtube-dl option.")
```
http://click.pocoo.org/6/options/#tuples-as-multi-value-options
http://click.pocoo.org/6/options/#multiple-options

This would allow the user to do something like this:
```
catt cast -y username borisjohnson -y password nohairdresser <url>
```
Click is then passing these value(s) on as a tuple containing one tuple for every ytdl option the user has set (we would of course have to convert "true"/"false" to the corresponding bools).

There's a few downsides to this approach:
+  If the user mistakenly supplies too many/few strings with one instance of ```-y```, ```catt``` will get very confused because other options or the url will be interpreted as a string belonging to ```-y```. There's no way to print a useful error message to the user when this happens.
+ Youtube-dl does not warn or fail, if the supplied options are wrong. Those are just ignored.
+ We would have to refer users to [youtube-dl source code](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/__init__.py#L315) for option names, as those sometimes differ from the ones given directly to the ```youtube-dl``` cli.

theychx:
@skorokithakis We could also just have a syntax like ```-y "{foo: bar, baz: ban}"```, and then run that through ```ast.literal_eval```. Not as pretty as my first suggestion, but much easier to make robust.

skorokithakis:
@theychx If we're going to do the latter, we might as well request JSON in the cli, but that gets a bit unwieldy. I wonder if it would be easier to drop the library dependency and just shell out to `youtube-dl` from the CLI. Then we could pass the options as they are, but I'm afraid that might be a bit less robust.

theychx:
> I wonder if it would be easier to drop the library dependency and just shell out to youtube-dl from the CLI

That has pretty far reaching consequences, so I'm not a fan of that.

skorokithakis:
Yeah, we'd have to do a bunch of work, so I'm not crazy about it.

xybydy:
I think @theychx's multiple tuples solution fits better. We could map the the youtube-dl's parameters into a dictionary to convert values into related data types before passing the parameters to youtube-dl.  With this approach, by not passing unmapped parameters to youtube-dl, we would have the control to prevent unexpected behavior.

skorokithakis:
I just realized I never replied to this. I agree with the tuples suggestion, as it was detailed in `catt cast -y username borisjohnson -y password nohairdresser <url>`. That seems like the best compromise to me.

theychx:
> If the user mistakenly supplies too many/few strings with one instance of -y, catt will get very confused because other options or the url will be interpreted as a string belonging to -y. There's no way to print a useful error message to the user when this happens.

Sorry for the change of heart, but I really don't think that this is acceptable. This will lead to confusing/irrelevant error messages. I still like that simple syntax though, but can't think of a way to make it robust.
We could make a ```click``` custom type and/or argument callback, in which we could verify the ytdl option. But that would also mean that we would have use some kind of delimiter (like ```-y "username|borisjohnson"```). We would also be able to determine whether the option is just a flag, thus enabling the user to just do: ```-y netrc```.
But again, that's a bit weird, so.....

skorokithakis:
Yes, that's why I'm partial to your solution only if it forces one `-y` per pair, to make it easier to tell if something is off. As you describe it, each `-y` only accepts a pair of parameters, so it won't be hard to visually inspect and make sure the pairs are correct (especially if they're quoted). All the other options seem to me either too complicated for the user, or too complicated for us (ie having to duplicate every single ytdl option).

theychx:
Hmm yeah, you're right. I'll just have to think of a way to improve on my initial proposal (without sacrificing the simple syntax).

theychx:
That could be implemented like this:
```
class YtdlOptParamType(click.ParamType):
    def convert(self, value, param, ctx):
        try:
            delimpos = value.index("=")
        except ValueError:
            self.fail("%s is not a valid key value pair." % value, param, ctx)

        ykey, yval = value[:delimpos], value[delimpos + 1:]
        yval = {"true": True, "false": False}.get(yval.lower(), yval)
        return (ykey, yval)

YTDL_OPT = YtdlOptParamType()
```
And then for ```cli.cast```:
```
@click.option("-y", "--ytdl-option", type=YTDL_OPT, multiple=True, metavar="OPT", help="Youtube-dl option.")
```

skorokithakis:
That seems reasonable to me, you can also do `if "=" not in value: raise Error; ykey, yval = value.split("=", 1)`

theychx:
Cheers. I've updated my suggestion.

skorokithakis:
Err, sorry, approved by mistake. Your suggestion sounds ideal to me, @theychx. @xybydy would you like to update the PR so we can merge?

catt/stream_info.py
```diff
@@ -31,14 +31,19 @@ class StreamInfoError(Exception):
...
             self.port = random.randrange(45000, 47000)
             self.is_local_file = True
         else:
-            self._ydl = youtube_dl.YoutubeDL({"quiet": True, "no_warnings": True})
+            self.ytdl_params = dict()
+            if ytdl_option:
+                self.ytdl_params = dict(ytdl_option)
+            self.ytdl_params.update({"quiet": True, "no_warnings": True})
+
+            self._ydl = youtube_dl.YoutubeDL(self.ytdl_params)
```
theychx: When the user specifies credentials for ```youtube-dl``` to use, it does not fail if those credentials are wrong/inadequate, so the current options are not really desirable in this case (I'm sure this is also the case for other options). I would prefer something like this:
```
DEFAULT_YTDL_OPTS = {"quiet": True, "no_warnings": True}
self._ydl = youtube_dl.YoutubeDL(dict(ytdl_option) if ytdl_option else DEFAULT_YTDL_OPTS)
```

skorokithakis: +1

catt/controllers.py
```diff
@@ -70,6 +70,8 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None):
                        controller is bypassed, and the one specified here is
                        returned instead.
     :type controller: str
+    :param ytdl_option: YoutubeDL options
```
skorokithakis: This should be a bit more descriptive, something like `Pairs of options to be passed to YoutubeDL.`

catt/cli.py
```diff
@@ -176,12 +189,15 @@ def process_subtitle(ctx, param, value):
 @click.option(
     "--no-subs", is_flag=True, default=False, help="Don't try to load subtitles automatically from the local folder."
 )
+@click.option("-y", "--ytdl-option", type=YTDL_OPT, multiple=True, metavar="OPT", help="Youtube-dl option.")
 @click.pass_obj
-def cast(settings, video_url, subtitle, force_default, random_play, no_subs):
+def cast(settings, video_url, subtitle, force_default, random_play, no_subs, ytdl_option):
```
skorokithakis: Can you rename `ytdl_option` to `ytdl_options` everywhere *except* this function? When the user passes it, it's one option, but when we receive it (i.e. from `setup_cast` onwards), it's going to be a tuple of multiple options, so it would be more accurate to say "options".

catt/stream_info.py
xybydy: But with this approach the default parameters will be replaced unless they're explicitly written when the user pass the other ytdl parameters. So how about something like this:
```python
    ytdl_params = {"quiet": True, "no_warnings": True}
            if ytdl_options:
                ytdl_params.update(ytdl_options)
            self._ydl = youtube_dl.YoutubeDL(ytdl_params)
```

catt/controllers.py
xybydy: Updating it as below if it's ok
```python 
ytdl_options: Pairs of options to be passed to YoutubeDL. For the available options to please refer to
                        https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L138-L317
```

skorokithakis: Ah, yes, that's great. Can you wrap the sentence a bit more so it looks tidier? Obviously the URL can't wrap, but that's okay.

catt/cli.py
xybydy: We also have not mentioned anywhere how pass youtube-dl  parameters, so I have updated the options short help text with *"Youtube-dl option. Should be passed as `key=value` format"*

catt/controllers.py
xybydy: Divided it to three lines, it's better now.

catt/cli.py
skorokithakis: Good call, although you might want to specify `-y key=value` so people don't think it's `--key=value` or something similar.

catt/stream_info.py
theychx: > But with this approach the default parameters will be replaced unless they're explicitly passed with the other ytdl parameters

That is intentional. Like I said, ```{"quiet": True, "no_warnings": True}``` is not desirable when used with credentials and probably many other options.

xybydy: These options increase verbosity level, that's all. You can get any error messages with or without these parameters. Especially without the `quiet` option, the use gets any stdout of ytdl prints.

theychx: Eh? When ```--no-warnings``` are set, ```WARNING: Unable to login: Invalid password``` is not printed.

xybydy: Youtube raises an error message when it's unable to provide the stream link. So it does not do anything with warning messages. I think there's no need for warning messages unless we want more verbosity for debugging.

The error messages I get with default options;
`ERROR: Unable to login, incorrect username and/or password`

`ERROR: Unable to download webpage: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:841)> (caused by URLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:841)'),))
Error: Remote resource not found.`

xybydy: No, I have just passed  wrong username and password

theychx: What url did you use?

xybydy: Well, I got SSL failure dailymotion, vk and youtube. For login error, I've used vk.com

theychx: OK, I'll have a look, after you've pushed the other changes.
(I still stand by my request though, as many options could alter the behaviour of ```catt```, and thus the extra verbosity will be useful to users.)

theychx: @xybydy What is your setup? I cannot reproduce the failing behaviour. I only get warnings (using ```youtube-dl 2018.8.4```).

theychx: I have no idea what's going on. @skorokithakis can you test this PR?

xybydy: Got it recorded at https://asciinema.org/a/5MDij53wAuCe709Wq7COJM5At

theychx: Thanks, I can reproduce the error with your vk.com link. But with youtube it's only warnings.

theychx: ..using ```https://www.youtube.com/feed/subscriptions```. It might be different with personal videos.

xybydy: As I've mentioned, it seems youtube-dl raises an error when ytdl is unable to provide a stream link, not 100% sure about that though. It seems, if the wrong credentials have nothing to do with providing the stream link, only warnings show up.

theychx: I still prefer ```dict(ytdl_option) if ytdl_option else DEFAULT_YTDL_OPTS```, as the extra verbosity is bound to be useful with a lot of these options.

skorokithakis: I agree with @theychx, as it will be surprising behaviour for the user if they specify options X and Y and we're secretly setting options A and B behind their back. When the user hasn't specified a preference, we can do whatever we want, but when they say they want a specific set of options, we shouldn't set anything they didn't explicitly request.

skorokithakis:
Looks great, thank you very much for your help and contribution!

PR #124, opened 2018-09-07
Minor image casting / Dashcast fixes

theychx:
A few fixes, so we have correct behaviour of ```status``` when casting images. Also ```DashCastController``` now fails with the control commands it doesn't support.


skorokithakis:
Looks good to me!

PR #125, opened 2018-09-08
Make save files portable

theychx:
Currently, an entry in a save file is tied to a specific device, which means that the save file can't be used elsewhere. I've tried to fix that, while keeping backwards compability with existing save files.
We still use real device names in ```~/.config/catt/state.json```.


skorokithakis:
Looks good, thank you!

PR #126, opened 2018-09-09
Very late spring cleaning

theychx:
:dancing_women:


skorokithakis:
Did `black` not take care of the white space after the docstring?

theychx:
Nope. I'm surprised it didn't complain.

theychx:
Did you see https://github.com/ambv/black/issues/355?

skorokithakis:
I had not, thank you. Hopefully Black will get an opinion about this soon and then we can just upgrade it.

PR #127, opened 2018-09-10
[WIP] Use new Youtube controller

theychx:
..as featured in ```PyChromecast 2.3.0```.

The new controller is both faster and more stable than the one we are currently using. It's missing ```clear_queue``` functionality though, and thus this is still WIP.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -274,6 +274,16 @@ def add(settings, video_url, play_next):
     cst.add(stream.video_id, play_next=play_next)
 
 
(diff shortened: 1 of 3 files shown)


skorokithakis:
Hey, do you use PyCharm, and if yes, do you have a license for it?

theychx:
Nope, I use vscode (and do not have a PyCharm license).

skorokithakis:
Ah, if you want one, I can give you one.

theychx:
That's very generous of you, but I don't feel like this is the time for me to switch editor.

PR #128, opened 2018-09-17
Switch from netifaces to ifaddr for get_local_ip

theychx:
```zeroconf``` have switched from ```netifaces``` to ```ifaddr``` (which, unlike ```netifaces```, is pure python), thus this PR.

```ifaddr``` is very simple though, and does not offer a means of identifying what interface the system considers to be the default, so I've attempted to implement something that returns the same(ish). Seems to work on Linux, but am not shure about macOS and Windows.

### DIFF
--- a/catt/stream_info.py
+++ b/catt/stream_info.py
@@ -1,3 +1,4 @@
+import ipaddress
 import random
 from pathlib import Path
 
@@ -32,10 +33,10 @@ class StreamInfoError(Exception):
 
 
 class StreamInfo:
-    def __init__(self, video_url, model=None, device_type=None, ytdl_options=None):
+    def __init__(self, video_url, host=None, model=None, device_type=None, ytdl_options=None):
         if "://" not in video_url:
             self._local_file = video_url
-            self.local_ip = self._get_local_ip()
+            self.local_ip = self._get_local_ip(host)
             self.port = random.randrange(45000, 47000)
             self.is_local_file = True
         else:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by theychx** [comment]
> We use this for the dev server, right?

Eh?

**Comment by skorokithakis** [comment]
Fair question, I don't know what I was thinking. We use this to discover our address for the *web* server to cast media to devices, right?

**Comment by theychx** [comment]
> We use this to discover our address for the web server to cast media to devices, right?

Yep.

Do you want to merge this now, and just wait and see what happens? Or should we make an effort to investigate whether this works on all of the 3 major platforms?

**Comment by skorokithakis** [comment]
I was wondering whether we can be a *bit* cleverer about this and try to guess which interface is the correct one, in the case where we get an IP for the Chromecast (which is true in the vast majority of cases). If a client has N interfaces, and the Chromecast is on `192.168.3.18`, we could pick the interface that is closest to `192.168.x.x`, so maybe `192.168.3.1` or `192.168.1.1`, or something like that, rather than pick `172.22.1.23`, which will probably not work.

We can merge this one and think about this in another PR, though. Do you want to remove "[WIP]" from the title? I have a personal rule to never merge PRs that say "WIP".

**Comment by theychx** [comment]
> Do you want to remove "[WIP]" from the title?

No, I like your suggestion, so I'll implement it.

**Comment by theychx** [comment]
Done

PS: You might want to know that ```zeroconf``` is now pulling in ```typing``` for py3.4.

File: catt/stream_info.py [hunk 1]
```diff
@@ -148,9 +149,19 @@ def set_playlist_entry(self, number):
...
             raise StreamInfoError("called on non-playlist")
 
-    def _get_local_ip(self):
-        interface = netifaces.gateways()["default"][netifaces.AF_INET][1]
-        return netifaces.ifaddresses(interface)[netifaces.AF_INET][0]["addr"]
+    def _get_local_ip(self, host):
+        for adapter in ifaddr.get_adapters():
+            for aip in adapter.ips:
+                try:
+                    ipaddress.IPv4Address(aip.ip)
```
skorokithakis: This looks great, except is there a reason to not support IPv6 here?

theychx: I figured it wouldn't be relevant on the users own network, but I can implement it if you prefer, no problem.

skorokithakis: I know a few people who run their internal network as IPv6-only, I just don't want to exclude them if it's easy to include them. Won't ifaddr return the IPv6 subnet in the same way as the IPv4 one? How much work is it to do this?

theychx: > Won't ifaddr return the IPv6 subnet in the same way as the IPv4 one? How much work is it to do this?

It will, it's the validating part that needs to change a bit. I'll fix it.

theychx: Not as easy as I thought, but should be dandy now.

skorokithakis:
This is fantastic, you're fantastic, everything is fantastic. Approved.

PR #133, opened 2018-10-11
fix: Give correct error messages with Dashcast

theychx:
Because of my earlier "fix", the wrong ```CastController``` was being returned if the Dashcast app was running on the cc. Resulting in irrelevant error messages when issuing a control command (which the dashcast ```CastController``` does not support).

The CI failure is still because of the DailyMotion extractor, which was fixed in the latest release of ```youtube-dl```. Any ideas? Or should we just update requirements?

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -129,9 +129,9 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None, ytdl_opt
 
     if app["app_name"] == "youtube":
         controller = YoutubeCastController(cast, app["app_name"], app["app_id"], prep=prep)
-    # We also check for controller, in the unlikely event that youtube-dl
-    # gets an extractor named "dashcast".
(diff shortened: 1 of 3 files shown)


theychx:
> Is there a downside to just updating requirements?

Nope, I was just wondering about the intricacies of package versions used in CI builds.
I'll make a seperate PR with the updated ```setup.py```.

skorokithakis:
We should update whenever we can, as long as it doesn't break anything. Unfortunately catt is almost impossible to automatically test...

theychx:
Indeed. Are you ok with this PR?

skorokithakis:
Yep, looks great!

theychx:
Merging is blocked for me when CI has failed, so I took the liberty of updating requirements here. For unknown reasons the dailymotion test still fails, so I gave up and replaced this test with a twitch one.

skorokithakis:
Looks great, thank you!

Issue #131, opened 2018-09-29
Typo in readme command write-config


KimOlsson:
On [readme](https://github.com/skorokithakis/catt) page it says that you need to run this line in order to write to config file:
`catt -d <name_of_chromecast> write_config`
**write-config** (with a dash) is how it probably is supposed to be, at least it is so on my Windows 10 installation.

Thank you very much for writing this program! It's awesome 👍

theychx:
You're right. Thank you for reporting this!

theychx:
@KimOlsson This was actually not a typo, but rather caused by a change in behaviour in one of the libraries we use. So the name is back to ```write_config```, as of ```0.8.1``` (just released).
Sorry for the confusion!

PR #134, opened 2018-10-12
fix: Correct cast_site and write_config command names

theychx:
..as ```click``` (7.0+) has begun replacing underscores with dashes in command names that are derived from function names.
This was the cause of the confusion in #131, something I didn't pick up on, as it's been ages since I've last used the ```write_config``` command.
You should consider making a release, to avoid further confusion.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,7 +78,7 @@ be selected, when you have not selected a device via the cli.
 
 You can write your choice of default device to ``catt.cfg`` by doing::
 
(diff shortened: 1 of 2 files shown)


skorokithakis:
Are there any other features in master that could block release? We've only made fixes since the last release, so this is a patch release, right?

theychx:
> so this is a patch release, right?

Yep, and it passes my testing scheme.

skorokithakis:
Released 0.8.1!

PR #136, opened 2018-10-23
Prep method fixes

theychx:
After a recent cc firmware upgrade, the behaviour of the cc after end of playback has changed if such a way, that the extra ```stream_type``` check in ```is_idle``` is no longer reliable. So I've resorted to a solution that Google can't break.
Also, I've introduced a new ```prep_info``` method so ```info``` and ```status``` commands are always available (as long as the cc is active).
Finally, I've sneaked in a slight improvement to url validation in ```cli.add```, in an attempt to weed out a few scenarios with irrelevant error msgs.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -268,7 +268,7 @@ def cast_site(settings, url):
 @click.pass_obj
 def add(settings, video_url):
     cst, stream = setup_cast(settings["device"], video_url=video_url, prep="control")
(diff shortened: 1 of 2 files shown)


skorokithakis:
Looks good! Do you want me to merge?

theychx:
Sure. (I don't mind doing it though :smiley: )

skorokithakis:
Go for it!

PR #137, opened 2018-10-23
Library catt

theychx:
Ehhmm dude, this one got away from me. One thing led to another, and I ended up with yet another redesign. My apologies for the less than readable diff.

You'll also notice that the code has even fewer docstrings than before. :rofl: 

But seriously, let's get the ball rolling with regards to how we can make the very-high level API as straight-forward as possible. I've added a simple example script you can look at.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -126,11 +126,14 @@ def setup_cast(device_name, video_url=None, controller=None, ytdl_options=None, 
 
     if controller:
         app_info = get_app_info(controller, cast_type, strict=True)
-    elif stream and prep == "app":
-        if stream.is_local_file:
-            app_info = get_app_info("default")
+    elif prep == "app":
+        if stream:
+            if stream.is_local_file:
+                app_info = get_app_info("default")
+            else:
+                app_info = get_app_info(stream.extractor, cast_type, show_warning=True if stream else False)
         else:
-            app_info = get_app_info(stream.extractor, cast_type, show_warning=True if stream else False)
+            app_info = get_app_info("default")
     else:
         app_info = get_app_info(cast.app_id, cast_type)
 
@@ -310,7 +313,6 @@ class MediaStatusListener:
     def __init__(self, current_state, states, invert=False, fail=False):
         if any(s not in VALID_STATE_EVENTS for s in states):
             raise ListenerError("invalid state(s)")
-
         if invert:
             self._states_waited_for = [s for s in VALID_STATE_EVENTS if s not in states]
         else:
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -511,6 +511,7 @@ class PlaybackBaseMixin:
     def play_playlist(self, playlist_id):
         raise NotImplementedError
 
+    @catch_namespace_error
     def wait_for(self, states, invert=False, fail=False):
         states = [states] if isinstance(states, str) else states
         media_listener = MediaStatusListener(
@@ -520,7 +521,8 @@ class PlaybackBaseMixin:
         media_listener.wait_for_states()
 
     def wait_for_playback_end(self):
-        self.wait_for(["BUFFERING", "PLAYING"], invert=True, fail=True)
+        self.wait_for("PLAYING")
+        self.wait_for(["BUFFERING", "PLAYING"], invert=True)
 
     def restore(self, data):
         raise NotImplementedError
@@ -592,7 +594,6 @@ class YoutubeCastController(CastController, MediaControllerMixin, PlaybackBaseMi
             for video_id in playlist[1:]:
                 self.add(video_id)
 
-    @catch_namespace_error
     def add(self, video_id):
         echo('Adding video id "%s" to the queue.' % video_id)
         self._prep_yt(video_id)
@@ -600,7 +601,6 @@ class YoutubeCastController(CastController, MediaControllerMixin, PlaybackBaseMi
         self.wait_for("BUFFERING", invert=True)
         self._controller.add_to_queue(video_id)
 
-    @catch_namespace_error
     def restore(self, data):
         self.play_media_id(data["content_id"])
         self.wait_for("PLAYING")
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -356,6 +356,31 @@ class CastController:
         elif prep == "info":
             self._prep_info()
 
+    def _prep_app(self):
+        """Make sure desired chromecast app is running."""
+
+        if not self._cast_listener.app_ready.is_set():
+            self._cast.start_app(self._cast_listener.app_id)
+            self._cast_listener.app_ready.wait()
+
+    def _prep_control(self):
+        """Make sure chromecast is not inactive or idle."""
+
+        self._check_inactive()
+        self._cast.media_controller.block_until_active(1.0)
+        if self._is_idle:
+            raise CattCastError("Nothing is currently playing.")
+
+    def _prep_info(self):
+        """Make sure chromecast is not inactive."""
+
+        self._check_inactive()
+        self._cast.media_controller.block_until_active(1.0)
+
+    def _check_inactive(self):
+        if self._cast.app_id == BACKDROP_APP_ID or not self._cast.app_id:
+            raise CattCastError("Chromecast is inactive.")
+
     @property
     def cc_name(self):
         return self._cast.device.friendly_name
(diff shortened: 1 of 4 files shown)


skorokithakis:
Jeeeez, what's with the /r/watchpeopledie example? The diff is pretty big, yep, I'll skim it and see if I have any comments and report back, thanks!

skorokithakis:
I can't say I'm in love with the API. In the simplest case, I'd prefer something like:

```
cc = Chromecast()
cc.play_url("https://whatever")
cc.wait()
```

Currently there's a lot of boilerplate just to play a URL, which we should probably abstract behind a nice, sane API and expose the less-frequent cases in other ways

theychx:
> Currently there's a lot of boilerplate just to play a URL

I agree, I just wanted to get things going before the PR got even more unreviewable. I'll log on to Gitter now.
> Jeeeez, what's with the /r/watchpeopledie example?

Yeah, using liveleak for this wasn't one of my grandest ideas ever.

catt/controllers.py
```diff
@@ -92,50 +63,81 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None, ytdl_opt
...
-    if controller:
-        if controller == "default":
-            app = DEFAULT_APP
-        else:
-            app = next(a for a in APP_INFO if a["app_name"] == controller)
-    elif stream and prep == "app":
-        if stream.is_local_file:
-            app = DEFAULT_APP
+def get_stream(url, device_info=None, host=None, ytdl_options=None):
+    if device_info:
```
skorokithakis: Minor suggestion, when initializing variables and then overwriting them on some condition (vs initializing them two different ways), I think this is slightly clearer:

```suggestion
    cc_info = cast_type = None
    if device_info:
        model_name = DEVICES_WITH_TWO_MODEL_NAMES.get(device_info.model_name, device_info.model_name)
        cc_info = (device_info.manufacturer, model_name)
        cast_type = device_info.cast_type
```

catt/controllers.py
```diff
@@ -92,50 +63,81 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None, ytdl_opt
...
+def get_app_info(id_or_name, cast_type=None, strict=False, show_warning=False):
+    if id_or_name == "default":
+        return DEFAULT_APP
+
+    field = "app_id" if re.match("[0-9A-F]{8}$", id_or_name) else "app_name"
+    try:
+        app_info = next(a for a in APP_INFO if a[field] == id_or_name)
+    except StopIteration:
+        if strict:
+            raise AppSelectionError("app not found (strict is set)")
```
skorokithakis:
```suggestion
            raise AppSelectionError("App not found (strict is set)")
```

catt/controllers.py
```diff
@@ -92,50 +63,81 @@ def setup_cast(device_name, video_url=None, prep=None, controller=None, ytdl_opt
...
+            if strict:
+                raise CattCastError(msg)
+            elif show_warning:
+                warning(msg)
+            app_info = DEFAULT_APP
+    return app_info
+
+
+def get_controller(cast, app_info, action=None, prep=None):
+    app_name = app_info["app_name"]
```
skorokithakis: This is a bit more obfuscated, but it should be clear enough, and the code becomes cleaner with multiple controllers:

```suggestion
    controller = {
        "youtube": YoutubeCastController,
        "dashcast": DashCastController,
    }.get(app_info["app_name"], DashCastController)
```

catt/controllers.py
```diff
@@ -164,6 +166,14 @@ class StateFileError(Exception):
     pass
 
 
+class ListenerError(Exception):
```
skorokithakis: We should probably put these exceptions into their own file, if we have more than two or three.

catt/controllers.py
```diff
@@ -300,24 +310,29 @@ def _is_app_ready(self, status):
...
-    def __init__(self, state):
-        self.not_buffering = threading.Event()
-        self.playing = threading.Event()
-        if state != "BUFFERING":
-            self.not_buffering.set()
-        if state == "PLAYING":
-            self.playing.set()
+    def __init__(self, current_state, states, invert=False, fail=False):
+        if any(s not in VALID_STATE_EVENTS for s in states):
+            raise ListenerError("invalid state(s)")
```
skorokithakis:
```suggestion
            raise ListenerError("Invalid state(s)")
```

catt/controllers.py
```diff
@@ -300,24 +310,29 @@ def _is_app_ready(self, status):
...
+        if invert:
+            self._states_waited_for = [s for s in VALID_STATE_EVENTS if s not in states]
+        else:
+            self._states_waited_for = states
+        if fail and current_state in self._states_waited_for:
+            raise ListenerError("condition is already met (fail is set)")
```
skorokithakis:
```suggestion
            raise ListenerError("Condition is already met (fail is set)")
```

catt/controllers.py
```diff
@@ -410,50 +448,32 @@ def _is_idle(self):
...
-        """Make sure chromecast is not inactive or idle."""
-
-        self._check_inactive()
-        self._cast.media_controller.block_until_active(1.0)
-        if self._is_idle:
-            raise CattCastError("Nothing is currently playing.")
-
-    def _prep_info(self):
-        """Make sure chromecast is not inactive."""
+    def volume(self, level):
```
skorokithakis: These are a bit misnamed, we should rename them to `set_volume`, `volume_up`, `volume_down` at some point.

catt/stream_info.py
```diff
@@ -53,9 +53,9 @@ def __init__(self, video_url, host=None, model=None, device_type=None, ytdl_opti
                 # We pop the "format" item, as it will make get_stream_info fail,
                 # if it holds an invalid value.
                 self._best_format = self._ydl.params.pop("format")
-            elif device_type in AUDIO_DEVICE_TYPES:
+            elif device_type and device_type in AUDIO_DEVICE_TYPES:
                 self._best_format = AUDIO_FORMAT
-            elif model in ULTRA_MODELS:
+            elif model and model in ULTRA_MODELS:
```
skorokithakis: What was the problem here that you had to add the `model and` part? I can't see any way for this to fail.

catt/controllers.py
theychx: I'm a bit uncertain on which case to use with exception messages that are not the ```click``` kind. Looking at exceptions from other contexts, additional messages seem to (mostly) start with lower-case.

catt/controllers.py
theychx: Cheers

catt/controllers.py
skorokithakis: Hm yeah. I generally prefer starting with uppercase, but I guess we can go either way, that's fine.

catt/controllers.py
theychx: I'll put that on my todo. Should all of our exceptions be in this file, regardless of flavour?

catt/controllers.py
theychx: Noted

catt/controllers.py
skorokithakis: Good question, probably not. For example, the ones in `controllers` should be different from the one that inherits from `ClickException` from the CLI. If we only have two here, we should leave them for now, it's just something to keep in the back of our mind for later.

catt/stream_info.py
theychx: You're right, I just like only going forward with the check if ```model``` actually holds a value (```None``` is a valid key in a python dict).

skorokithakis: Ah, right, that's why I was confused (because `None` isn't in `ULTRA_MODELS` so the check is not strictly necessary). Looks good!

catt/controllers.py
theychx: Cheers

theychx:
Why would flake8 only fail with py3.6?

skorokithakis:
I'm not sure, but it looks like an easy fix.

skorokithakis:
I don't know, but [the changelog](https://docs.python.org/3.6/whatsnew/3.6.html) clarifies:

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.

Non-valid escape sequences are deprecated, as they should be (because you thought your code was doing something but it was actually doing something else, i.e. trying to escape that sequence).

theychx:
Ah, thank you for that info.

PR #139, opened 2018-11-06
More api stuff

theychx:
The prototype is now a bit more complete. I've reimplemented the example.
I've also begun a quest to get rid of as many of the click exception raise statements as possible, now that ```controllers.py``` do not only have to cater to ```cli.py```.

### DIFF
--- a/catt/api.py
+++ b/catt/api.py
@@ -1,4 +1,4 @@
-from .controllers import get_app_info, get_cast, get_cast_with_ip, get_chromecasts, get_controller, get_stream
+from .controllers import get_app_info, get_cast_with_ip, get_chromecast, get_chromecasts, get_controller, get_stream
 
 
 def discover():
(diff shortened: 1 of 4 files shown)


skorokithakis:
Sorry, I was on a ten-day trip around Europe and just came back. I will look at this ASAP!

skorokithakis:
This looks good to me! I love the refactoring out of the exceptions, which made the code more modular. It would be nice if we could start using types so we can add mypy to our list of checks, but that's for later. Thanks!

theychx:
How much additional functionality would it take, for you to be able to reimplement your own scripts using this api?

skorokithakis:
My own scripts pretty much just need to find the devices by name and play a URL on them. I will try reimplementing them today if I have time and we can discuss my thoughts.

skorokithakis:
Alright, first of all, sorry for the delay (wow, exactly one month), I have been pretty busy with other things :(

I tried the API as it exists in the `api-develop` branch and I think it's perfect. It's super easy to get started with, sane, and supports all the use cases I needed. Do you think we can merge it into master, @theychx? I found it a pleasure to use.

theychx:
> sorry for the delay

No worries, I'd forgotten about this conversation myself.
> Do you think we can merge it into master?

I made a small test project that utilizes the api, where I unfortunately discovered some odd behaviour from cc's when they recieve a 404. The last ```player_state``` signal that is emitted when this happens is "BUFFERING". And that makes the blocking logic in ```play_url``` a bit tricky to get right (it would block forever if this is not adressed). So as a minimum, I'd like to fix this issue before merging.

skorokithakis:
Hmm, yeah, that sounds reasonable. Too bad about the problem, but you're right that we should fix it before merging.

PR #142, opened 2018-12-27
First attempt at fixing 404 issues

theychx:
Let's try this again :smile: , and then merge your changes into ```api-develop``` before doing the final merge.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -5,7 +5,7 @@ import tempfile
 import threading
 from enum import Enum
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional  # noqa
 
 import pychromecast
 from click import ClickException, echo
@@ -463,8 +463,8 @@ class CastController:
 
 
 class MediaControllerMixin:
-    _is_seekable: Any
-    _cast: pychromecast.Chromecast
+    _is_seekable = None  # type: Any
+    _cast = None  # type: pychromecast.Chromecast
 
     def play(self):
         self._cast.media_controller.play()
@@ -494,7 +494,7 @@ class MediaControllerMixin:
 
 
 class PlaybackBaseMixin:
-    _cast: pychromecast.Chromecast
+    _cast = None  # type: pychromecast.Chromecast
 
     def play_media_url(self, video_url: str, **kwargs) -> None:
         raise NotImplementedError
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -303,11 +303,14 @@ class MediaStatusListener:
             self._states_waited_for = [s for s in VALID_STATE_EVENTS if s not in states]
         else:
             self._states_waited_for = states
-        if fail and current_state in self._states_waited_for:
-            raise ListenerError("condition is already met (fail is set)")
 
         self._state_event = threading.Event()
         self._current_state = current_state
+        if self._current_state in self._states_waited_for:
+            if fail:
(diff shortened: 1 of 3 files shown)


catt/controllers.py
```diff
@@ -497,15 +499,15 @@ def play_media_id(self, video_id):
     def play_playlist(self, playlist_id):
         raise NotImplementedError
 
-    def wait_for(self, states, invert=False, fail=False):
+    def wait_for(self, states, invert=False, fail=False, timeout=None):
```
skorokithakis: Can you add some types to these function definitions? It'll make things clearer.

skorokithakis:
Looks good to me, some type declarations might be nice, but it's otherwise good to merge!

catt/controllers.py
theychx: How would I annotate params with multiple valid types, without the ```typing``` module?

skorokithakis: Hmm, would this not work?:

```
def wait_for(self, states, invert:bool=False, fail:bool=False, timeout:Optional[int]=None):
```

Also, what's wrong with using the `typing` module?

theychx: > Also, what's wrong with using the typing module?

It's py3.5+ only.

We could add it as a dependency for 3.4, as ```zeroconf``` is already doing this.

skorokithakis: Aw, I thought they added it from the start on 3. Yeah, that would be ideal, it's super useful. I can enable mypy checks afterwards.

catt/controllers.py
```diff
@@ -490,17 +494,18 @@ def skip(self):
 
 
 class PlaybackBaseMixin:
-    def play_media_url(self, video_url, **kwargs):
+    _cast: pychromecast.Chromecast
```
skorokithakis: Do we want to make these things class variables? That would break a whole slew of things.

theychx: It's for mypy, and not intended as a class var. But I must admit that I've kind of lost control of this PR, as I struggle to understand how to get typing to work across python versions. As I see it, I would have to use old-style type hints, as I get syntax errors with annotations such as these (w. py3.4).

skorokithakis: You get syntax errors because of the variable annotations, which are new. Just keep the function annotations and we're good, we don't need to go full Python 3.6 on this.

catt/controllers.py
```diff
@@ -460,23 +463,26 @@ def kill(self, idle_only=False):
 
 
 class MediaControllerMixin:
+    _is_seekable: Any
```
skorokithakis: Basically just removing these two lines and the `_cast` below is good.

skorokithakis: Yeah, that SO answer has the right idea:


```
_is_seekable = None  # type: Any
```

By the way, isn't this a `bool`?

theychx: > By the way, isn't this a bool?

Yes, but that doesn't work either. :sob:

skorokithakis: Hmm, it complains about the incompatibility, you mean?

theychx: Yep

skorokithakis: Yeah, I hate mypy sometimes... Anyway, this looks good now, and I haven't been able to get it to pass by changing the bool type, so I think we're good to merge.

theychx: Thanks!

skorokithakis: Thank you!

PR #147, opened 2018-12-30
Use glob in hunt_subtitle

theychx:
:fireworks:

### DIFF
--- a/catt/util.py
+++ b/catt/util.py
@@ -36,13 +36,13 @@ def guess_mime(path):
 def hunt_subtitle(video):
     """Searches for subtitles in the current folder"""
 
-    video_path = Path(video)
-    video_path_stem_lower = video_path.stem.lower()
-    for entry_path in video_path.parent.iterdir():
-        if entry_path.is_dir():
+    video_dir_path = Path(video).parent
+    for ext in ["*.vtt", "*.VTT", "*.srt", "*.SRT"]:
+        try:
+            sub = next(video_dir_path.glob(ext))
+        except StopIteration:
             continue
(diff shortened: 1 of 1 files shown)


catt/util.py
```diff
@@ -36,13 +36,13 @@ def guess_mime(path):
 def hunt_subtitle(video):
     """Searches for subtitles in the current folder"""
 
-    video_path = Path(video)
-    video_path_stem_lower = video_path.stem.lower()
-    for entry_path in video_path.parent.iterdir():
-        if entry_path.is_dir():
+    video_dir_path = Path(video).parent
+    for ext in ["*.vtt", "*.VTT", "*.srt", "*.SRT"]:
+        try:
```
skorokithakis: Is this not the same as:

```
for ext in ["vtt", "VTT", "srt", "SRT"]:
    subs = video_dir_path.glob("*." + ext)
    if not subs:
        continue
    return str(subs[0].resolve())
```

skorokithakis: I wish globs supported multiple extensions a la regexps, or at least case-insensitivity, but I didn't find anything.

theychx: > or at least case-insensitivity

I found this a bit annoying as well.

theychx: ```Path.glob``` returns a generator, thus the try/except clause.

skorokithakis: Err, yes, sorry, I'm missing a `list()` on my code, but I guess that would be less performant. This looks good, it just bugs me a bit that we're repeating the `*.` for each extension.

theychx: I already fixed that dude. Happy new year!

skorokithakis: Oh oops, so you did. Happy new year!

PR #148, opened 2019-01-08
Testing scheme

theychx:
Yo! Yo!
I've liberated my testing scheme (or some of it, at least) from the crappy shell script where it used to live. It's not entirely done, but works (and current master passes), so is good to merge.


realcc_tests/test_procedure.py
```diff
@@ -0,0 +1,200 @@
...
+        for test in suite:
+            test.set_cmd_base(cbase)
+            click.echo(test.desc + "  ->  ", nl=False)
+            if test.run():
+                click.secho("success!", fg="green")
+                test_outcomes.append(True)
+            else:
+                click.secho("failure!", fg="red")
+                test_outcomes.append(False)
+            if test.dump:
```
theychx: What is the most sane way of communicating that a test that is supposed to fail, actually fails? The success/failure options seems inadequate here.

skorokithakis: "Outcome as expected/outcome contrary to expected"? Also "test success/test failure" communicates that the test succeeded, even if the test was to check that something fails, in my opinion.

theychx: cheers

Issue #152, opened 2019-01-12
Minor API changes


skorokithakis:
Hey @theychx, I tried to use the API again and found a few omissions:

1. We should probably expose the `_cast` property, so the consumer can read the IP address, name, etc.
2. I think we should rename `ctrl` to `control` or `controller`, whichever it is.

What do you think?

skorokithakis:
Actually, I think we should proxy the properties and expose a better API for the `pychromecast` devices, i.e. add proxy properties for its fields onto the `CattDevice`.

theychx:
> I think we should proxy the properties

I agree. ```CattDevice``` already exposes name and ip-adress (```_create_cast``` updates those attributes with the data from the ```Chromecast``` object).

One of these days, I'll make a PR with the renaming of ```ctrl``` to ```control```. Any additional info from the ```Chromecast``` object you think would be useful as ```CattDevice``` attributes, I can include such changes in that PR.

skorokithakis:
I'm just about to push it now, although I missed the exposed name and IP, looking for those now.

PR #153, opened 2019-01-12
Rename some variables

skorokithakis:


skorokithakis:
I've renamed some things, I thought `ip_addr` looks more Pythonic with an underscore between the words, so I renamed that.

What do you think about creating getters for `name`, `ip_addr` and `uuid`? That will make it much easier to document, although I'm not sure if we can document instance variables with things like Sphinx. If we can, we should do it that way.

skorokithakis:
Closes #152, by the way.

theychx:
> What do you think about creating getters for name, ip_addr and uuid?

I would prefer ```@property``` decorated methods.

skorokithakis:
> I would prefer `@property` decorated methods.

That's what I meant by getters, vs just `self.name = self._cast.name`.

catt/api.py
```diff
@@ -11,56 +11,60 @@
...
+        self.ip_addr = self._cast.host
+        self.uuid = self._cast.uuid
 
-    def _create_ctrl(self) -> None:
+    def _create_controller(self) -> None:
         self._cast_controller = get_controller(self._cast, get_app_info("default"))
 
     @property
-    def ctrl(self):
+    def controller(self):
```
theychx: I would prefer ```control``` over ```controller```, as it's more distinct from the slightly confusing controller/castcontroller "paradigm".

skorokithakis: `control` is a verb though, implying a function. Or does it return "a control"?

theychx:
Ah, ok. No problem then.

skorokithakis:
There's no problem with that, right? It doesn't fetch the name through the network or something?

catt/api.py
theychx: > Or does it return "a control"?

That is what I meant, I see your point though, about it being a verb. I just dont' like ```controller```.

skorokithakis: Yeah... We can find a completely different name, it doesn't have to be `controller`.

theychx:
> It doesn't fetch the name through the network or something?

Nope, only the one time the ```Chromecast``` object is created.

skorokithakis:
Thought so, so we're good to make these properties. I think we're good to leave it for now and only change it when it becomes a documentation tooling problem.

skorokithakis:
Merged, thanks!

theychx:
Thank you!

theychx:
> Closes #152, by the way.

Github magic didn't close the issue you opened. Is it the comma? Probably just a gh brainfart.

PR #155, opened 2019-01-16
Small cleanup

theychx:
Howdy
I've moved a bit more stuff to ```util.py```.
Also, I have returned ```hunt_subtitle``` to its original state, as my new implementation was lame and did not work as intended. Because of the casesensitivity problem of ```Path.glob```, I couldn't figure out how to fix my code in a sane way (sorry about not testing this properly).

Finally, current master is in a release-worthy state, so please consider making a release. One thing to note though, is that I intend to rework exceptions, which might lead to changes that break the "stable" API.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -12,7 +12,7 @@ import requests
 
 from .controllers import Cache, CastState, StateFileError, StateMode, get_chromecast, get_chromecasts, setup_cast
 from .http_server import serve_file
-from .util import convert_srt_to_webvtt, convert_srt_to_webvtt_helper, hunt_subtitle, warning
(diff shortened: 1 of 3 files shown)


skorokithakis:
This looks good to me. When you say "stable API", do you mean the API we just added? That would be fine, since we haven't actually released it yet.

theychx:
> That would be fine, since we haven't actually released it yet.

Yes exactly, which is why I mentioned it in the context of making a new release.
I have no intention of reworking anything in ```api.py```, only fix bugs and augment it in a non-breaking way. It's just that if we rename a lot of exceptions that could be thrown at users of the API, we are strictly speaking breaking it (or?).

skorokithakis:
Yeah, as long as it's not documented, I consider it unreleased. Changing exceptions definitely counts as breaking the API, but it's not out yet, so break away.

PR #157, opened 2019-01-18
Relax url validation in add cmd a bit

theychx:
..so it is possible to add a video id from a mixed video/playlist url to the queue.

Please consider making a release. The ability to cast a yt playlist without the lame procedure of adding video ids one by one is pretty neat.


skorokithakis:
I don't know, it looks like the more I wait the more useful stuff you keep adding 😛

theychx:
I promise I'll keep adding useful stuff, even after you've made a new release. :smiley:

skorokithakis:
0.9.0 released uneventfully!

PR #158, opened 2019-01-20
Use namedtuples instead of dicts for app_info

theychx:
:tv:

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -91,11 +91,9 @@ def get_app_info(id_or_name, cast_type=None, strict=False, show_warning=False):
     if id_or_name == "default":
         return DEFAULT_APP
 
+    arg_is_id = True if re.match("[0-9A-F]{8}$", id_or_name) else False
     try:
-        if re.match("[0-9A-F]{8}$", id_or_name):
-            app_info = next(a for a in APP_INFO_LIST if a.id == id_or_name)
-        else:
-            app_info = next(a for a in APP_INFO_LIST if a.name == id_or_name)
+        app_info = next(a for a in APP_INFO_LIST if (a.id if arg_is_id else a.name) == id_or_name)
     except StopIteration:
         if strict:
             raise AppSelectionError("app not found (strict is set)")
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -31,7 +31,7 @@ class App:
         self.supported_device_types = supported_device_types
 
 
-APP_INFO_LIST = [
+APPS = [
     App(app_name="youtube", app_id="233637DE", supported_device_types=["cast"]),
     App(app_name="dashcast", app_id=DASHCAST_APP_ID, supported_device_types=["cast", "audio"]),
 ]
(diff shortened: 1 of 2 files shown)


theychx:
I was bored, so I made a preemptive strike on your review. :smile:

skorokithakis:
Haha, strike successful! By the way, I wouldn't be opposed to pulling in [attrs](https://github.com/python-attrs/attrs) as a dependency at any time if you think it would be useful.

catt/controllers.py
```diff
@@ -528,9 +536,7 @@ class DefaultCastController(CastController, MediaControllerMixin, PlaybackBaseMi
     def __init__(self, cast, name, app_id, prep=None):
```
skorokithakis: These arguments look suspiciously like we're passing an `app: App` in here, should we just pass that object?

theychx: Nice catch. Let me fix that.

catt/controllers.py
```diff
@@ -94,13 +94,13 @@ def get_stream(url, device_info=None, host=None, ytdl_options=None):
...
 
-def get_app_info(id_or_name, cast_type=None, strict=False, show_warning=False):
+def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
     if id_or_name == "default":
         return DEFAULT_APP
 
     arg_is_id = True if re.match("[0-9A-F]{8}$", id_or_name) else False
     try:
-        app_info = next(a for a in APP_INFO_LIST if (a.id if arg_is_id else a.name) == id_or_name)
+        app = next(a for a in APPS if (a.id if arg_is_id else a.name) == id_or_name)
```
skorokithakis: The generator/`next` thing here might be a bit too confusing to save space on a list of two elements, but that's what we had before, so it's fine.

catt/controllers.py
```diff
@@ -337,14 +336,14 @@ def wait_for_states(self, timeout=None):
 
 
 class CastController:
-    def __init__(self, cast, name, app_id, prep=None):
+    def __init__(self, cast, app, prep=None):
```
skorokithakis: Can we add types here?

catt/controllers.py
theychx: I actually had an if clause with a generator for each argument type initially. Would you prefer that?

skorokithakis: We're just trying to find the app with the given id, right? Why not just a for/else loop with a return?

catt/controllers.py
theychx: Let's see if Miss Mypy has one of her good days.

skorokithakis: :crossed_fingers:

catt/controllers.py
theychx: Or with the given name. I brewed those together to make calls to the function cleaner.

skorokithakis: Ah okay, I'm not fussy about it. I would have done something like:

```
for app in APPS:
    if id_or_name in (app.id, app.name):
        return app
else:
    return DEFAULT_APP  # todo strict handling etc
```

Which I think looks a bit simpler.

catt/controllers.py
```diff
@@ -85,40 +94,38 @@ def get_stream(url, device_info=None, host=None, ytdl_options=None):
     return StreamInfo(url, host=host, model=cc_info, device_type=cast_type, ytdl_options=ytdl_options)
 
 
-def get_app_info(id_or_name, cast_type=None, strict=False, show_warning=False):
+def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
     if id_or_name == "default":
```
skorokithakis: Do we need this? If `id_or_name` is `"default"` the code below is going to return the default app anyway, no?

catt/controllers.py
theychx: I don't like "for loop else". I think it looks weird.

catt/controllers.py
theychx: True. I think I just put it there to not engage in any logic if the requester specifically needs "default".

catt/controllers.py
theychx: Mypy doesn't like me running ```register_handler``` on an attribute that might not be there, so I made that error go away.

theychx:
Any further comments before I merge?

catt/controllers.py
skorokithakis: You can just not use the else. It'll return if it matches in the loop.

catt/controllers.py
skorokithakis: Sounds good!

catt/controllers.py
skorokithakis: Sounds good to me.

skorokithakis:
Nope, looks good to me!

PR #161, opened 2019-01-22
Fix and simplify get_app

theychx:
I incorporated half of your suggestion into this PR. :smiley: I also fixed a bug introduced in the previous PR.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -30,11 +30,12 @@ class App:
         self.supported_device_types = supported_device_types
 
 
+DEFAULT_APP = App(app_name="default", app_id="CC1AD845", supported_device_types=["cast", "audio", "group"])
 APPS = [
+    DEFAULT_APP,
     App(app_name="youtube", app_id="233637DE", supported_device_types=["cast"]),
     App(app_name="dashcast", app_id=DASHCAST_APP_ID, supported_device_types=["cast", "audio"]),
 ]
-DEFAULT_APP = App(app_name="default", app_id="CC1AD845", supported_device_types=["cast", "audio", "group"])
 
 
 def get_chromecasts():
@@ -97,7 +98,7 @@ def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
     try:
         app = next(a for a in APPS if id_or_name in [a.id, a.name])
     except StopIteration:
-        if strict and id_or_name != "default":
+        if strict:
             raise AppSelectionError("app not found (strict is set)")
         else:
             return DEFAULT_APP
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -1,6 +1,5 @@
 import hashlib
 import json
-import re
 import tempfile
 import threading
 from enum import Enum
@@ -95,11 +94,10 @@ def get_stream(url, device_info=None, host=None, ytdl_options=None):
 
 
 def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
-    arg_is_id = True if re.match("[0-9A-F]{8}$", id_or_name) else False
     try:
-        app = next(a for a in APPS if (a.id if arg_is_id else a.name) == id_or_name)
+        app = next(a for a in APPS if id_or_name in [a.id, a.name])
     except StopIteration:
-        if strict:
+        if strict and id_or_name != "default":
             raise AppSelectionError("app not found (strict is set)")
         else:
             return DEFAULT_APP


catt/controllers.py
```diff
@@ -95,11 +94,10 @@ def get_stream(url, device_info=None, host=None, ytdl_options=None):
 
 
 def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
-    arg_is_id = True if re.match("[0-9A-F]{8}$", id_or_name) else False
     try:
-        app = next(a for a in APPS if (a.id if arg_is_id else a.name) == id_or_name)
+        app = next(a for a in APPS if id_or_name in [a.id, a.name])
     except StopIteration:
-        if strict:
+        if strict and id_or_name != "default":
```
theychx: The simplification of logic in the previous PR introduced a bug. I have a test (as featured in ```test_procedure.py```) that casts low resolution content from twitch using ```--ytdl-opt```. In this case ```strict``` would be set, so we would fail, as "default" is being specifically requested.

skorokithakis: Ah, I see. Is there a reason why we aren't including the default app in the app list?

Also, I guess this is why you had the short-circuit in the logic, but there was no accompanying comment, so we forgot about the reason. That's a lesson in always leaving comments for non-obvious functionality.

catt/controllers.py
```diff
@@ -95,11 +94,10 @@ def get_stream(url, device_info=None, host=None, ytdl_options=None):
 
 
 def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
-    arg_is_id = True if re.match("[0-9A-F]{8}$", id_or_name) else False
     try:
-        app = next(a for a in APPS if (a.id if arg_is_id else a.name) == id_or_name)
+        app = next(a for a in APPS if id_or_name in [a.id, a.name])
```
skorokithakis: A tuple is a slightly more lightweight alternative to the list here, but I'm not fussed either way.

catt/controllers.py
skorokithakis: It looks like we only use `APPS` in `get_app`, but then special-case `DEFAULT_APP` by not including it in the list. The code would be simpler if we did, no?

theychx: > That's a lesson in always leaving comments for non-obvious functionality

Noted. :smiley:

theychx: > It looks like we only use APPS in get_app, but then special-case DEFAULT_APP by not including it in the list. The code would be simpler if we did, no?

True. The problem is that we are using ```DEFAULT_APP``` later in ```get_app``` and then also in [```DefaultCastController```](https://github.com/skorokithakis/catt/blob/master/catt/controllers.py#L535).
We could of course have our cake and eat it by including the default app in ```APPS```, and then do ```DEFAULT_APP = next(a for a in APPS if a.name ==  "default")```.

theychx: (or we could change the datastructure again :smile: )

theychx:
Thanks!

theychx:
CI tests are failing because of changes to format selection stuff in the newest youtube-dl release. I am currently investigating this.

skorokithakis:
Looks good! Although, why are tests failing?

theychx:
The format selection syntax actually changed. The man file was also changed, so it is not a bug.
I can submit another PR with the fix, and then we can roll out a new release.

theychx:
We should just update the youtube-dl requirement to the latest release, so people can fix the issue by upgrading everything.

skorokithakis:
I agree, but that doesn't guarantee that it won't break again when they change things next time. The problem with youtube-dl is that they don't use semantic versioning so we can't pin to a minor version and be sure the API won't change under us, so right now there's nothing we can do about this, unfortunately.

theychx:
> but that doesn't guarantee that it won't break again when they change things next time

Hasn't that been a fundamental premise for this project all along?
> they don't use semantic versioning

That is a rather academic point, as youtube-dl slowly becomes useless without updated extractors.

Anyway, this wasn't as easy a fix as I thought, so I suggest that we pin to ```2019.1.17```, and then make a new release.

skorokithakis:
> Hasn't that been a fundamental premise for this project all along?

How do you mean? I, for one, don't expect old versions to break suddenly.

> That is a rather academic point, as youtube-dl slowly becomes useless without updated extractors.

Yeah, I guess that's why pinning it doesn't really work well. They could have at least split the extractors  into some other module via a simple interface, so we could upgrade the two separately, but it is what it is.

> I suggest that we pin to 2019.1.17, and then make a new release.

I agree, but this will only last until the extractors start breaking. Still, it'll buy us some time.

PR #169, opened 2019-02-26
Exceptions rework

theychx:
Heyhey
Here's my proposal for a more useful organizing of exceptions. Not perfect, but definitely an improvement IMO.
I've had to implement the mechamism that transforms exceptions into cli error messages myself, now that we don't have the convenience of ```click.ClickException``` at our disposal anymore (which I have ditched, as the main idea here is that all of our own exceptions inherit from the same base class).
I've also had to slightly rework some of the subtitles stuff, to make it fit the ideas of this PR.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -6,13 +6,13 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Optional  # noqa
 
-import click
 import pychromecast
 from pychromecast.controllers.dashcast import APP_DASHCAST as DASHCAST_APP_ID
 from pychromecast.controllers.dashcast import DashCastController as PyChromecastDashCastController
 from pychromecast.controllers.youtube import YouTubeController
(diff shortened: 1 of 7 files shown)


catt/controllers.py
```diff
@@ -99,19 +99,19 @@ def get_app(id_or_name, cast_type=None, strict=False, show_warning=False):
...
+            raise AppSelectionError("App not found (strict is set)")
         else:
             return DEFAULT_APP
 
     if app.name == "default":
         return app
 
     if not cast_type:
-        raise AppSelectionError("cast_type is needed for app selection")
+        raise AppSelectionError("Cast type is needed for app selection")
```
skorokithakis: What, no final period here? :stuck_out_tongue:

skorokithakis: Oh you're removing, never mind.

PR #171, opened 2019-03-02
Use app id constants from pychromecast.config instead of our own

theychx:
:tv:

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -7,7 +7,10 @@ from pathlib import Path
 from typing import Any, Optional  # noqa
 
 import pychromecast
-from pychromecast.controllers.dashcast import APP_DASHCAST as DASHCAST_APP_ID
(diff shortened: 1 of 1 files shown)


catt/controllers.py
```diff
@@ -7,7 +7,10 @@
 from typing import Any, Optional  # noqa
 
 import pychromecast
-from pychromecast.controllers.dashcast import APP_DASHCAST as DASHCAST_APP_ID
+from pychromecast.config import APP_BACKDROP as BACKDROP_APP_ID
+from pychromecast.config import APP_DASHCAST as DASHCAST_APP_ID
+from pychromecast.config import APP_MEDIA_RECEIVER as MEDIA_RECEIVER_APP_ID
+from pychromecast.config import APP_YOUTUBE as YOUTUBE_APP_ID
```
theychx: ```isort``` doesn't like me doing these imports in one statement, for some reason.

PR #172, opened 2019-03-02
Add -h alias for --help option

theychx:
Something that has annoyed me for a long time, but I was to lazy to fix.
Also works with cmd help!


skorokithakis:
Jeez, I can't believe this was missing. Thanks!

PR #187, opened 2019-03-25
Improvements to test_procedure

theychx:
I've also added some local file casting tests to what remains of my old testing scheme, just in case :smiley: 

I've run both the old and the new testing schemes, and we should be ready for release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixes
 
+* Refuse to install under Python 2. [Stavros Korokithakis]
+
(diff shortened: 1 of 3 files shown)


skorokithakis:
This looks great, I just have one question that I never realized until now: What's the `sleep` parameter for?

theychx:
> What's the sleep parameter for?

It's just so, after a `catt` command in a test has been executed we wait for some seconds before verifying the test value, to make sure that the test doesn't fail for irrelevant reasons (like local setup).

PR #197, opened 2019-06-13
feat: Add ability to use ip-address as device argument 

### LINKED ISSUE #196: Using IP of chromecast as alternative to device name
Hi,

In some case, it's easier to use IP than device name. Seeing the code a simple alternative in the controllers must make this possible.
Do you think it could be added ?

theychx:
Howdy! :cowboy_hat_face: 

This won't work with audio device groups, as the "group leader" is on a different port.
It should be easy to add a `--port` option. I'll implement that as soon as I get my hands on another cc audio, so I can test.

fixes #196

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -14,7 +14,7 @@ from . import __version__
 from .controllers import Cache, CastState, StateFileError, StateMode, get_chromecast, get_chromecasts, setup_cast
 from .error import CastError, CattUserError, CliError, SubsEncodingError
 from .http_server import serve_file
-from .util import convert_srt_to_webvtt, echo_json, human_time, hunt_subtitle, read_srt_subs, warning
+from .util import convert_srt_to_webvtt, echo_json, human_time, hunt_subtitle, is_ipaddress, read_srt_subs, warning
 
 CONFIG_DIR = Path(click.get_app_dir("catt"))
 CONFIG_PATH = Path(CONFIG_DIR, "catt.cfg")
@@ -73,11 +73,11 @@ def process_path(ctx, param, value):
     return path
 
 
-def get_device(ctx, param, value):
-    try:
-        return ctx.default_map["aliases"][value]
-    except KeyError:
+def process_device(ctx, param, value):
+    if is_ipaddress(value):
         return value
+    else:
+        return ctx.default_map["aliases"].get(value, value)
 
 
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
@@ -85,7 +85,7 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
 @click.group(context_settings=CONTEXT_SETTINGS)
 @click.option("--delete-cache", is_flag=True, help="Empty the Chromecast discovery cache.")
-@click.option("-d", "--device", metavar="NAME", callback=get_device, help="Select Chromecast device.")
+@click.option("-d", "--device", metavar="NAME_OR_IP", callback=process_device, help="Select Chromecast device.")
 @click.version_option(version=__version__, prog_name="catt", message="%(prog)s v%(version)s, Winter Waterfall.")
 @click.pass_context
 def cli(ctx, delete_cache, device):
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -74,6 +74,13 @@ def process_path(ctx, param, value):
 
 
 def process_device(ctx, param, value):
+    """
+    Resolve real device name when value is an alias.
+
+    :param value: Can be an ip-address or a name (alias or real name).
+    :type value: str
+    """
+
     if is_ipaddress(value):
         return value
     else:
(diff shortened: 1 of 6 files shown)


skorokithakis:
Hello!

This is great! I was thinking we'd try with the address wholesale and see if that failed, and then try with the name, but maybe that has unacceptable latency. This way prevents people from using hostnames like `livingroom.myhouse.local`, but I guess that's a pretty rare use case.

catt/cli.py
```diff
@@ -73,19 +73,19 @@ def process_path(ctx, param, value):
     return path
 
 
-def get_device(ctx, param, value):
-    try:
-        return ctx.default_map["aliases"][value]
-    except KeyError:
+def process_device(ctx, param, value):
```
skorokithakis: This might need a docstring and some better naming of variables, since `value` doesn't tell me what the value should be.

catt/cli.py
```diff
@@ -73,19 +73,19 @@ def process_path(ctx, param, value):
...
+    if is_ipaddress(value):
         return value
+    else:
+        return ctx.default_map["aliases"].get(value, value)
```
skorokithakis: Maybe the aliases should have higher priority? I can't see them conflicting with IP addresses, but just a thought.

catt/controllers.py
```diff
@@ -69,20 +69,27 @@ def get_chromecast_with_ip(device_ip, port=DEFAULT_PORT):
         return None
 
 
-def get_cast(device_name=None):
+def get_cast(device=None):
```
skorokithakis: This could use a short docstring as well.

catt/cli.py
theychx: You're right, I just don't like checking ip-addresses against the alias list.

skorokithakis: That's fair.

theychx:
> I was thinking we'd try with the address wholesale and see if that failed, and then try with the name, but maybe that has unacceptable latency. This way prevents people from using hostnames like livingroom.myhouse.local, but I guess that's a pretty rare use case

Please explain in more detail.

catt/cli.py
theychx: I like all the callback functions being similar in this regard. But let me add the docstring.

catt/controllers.py
theychx: Will do.

skorokithakis:
Basically just try to talk to the Chromecast by IP and if that fails (the value is an IP but nothing's there, it's a hostname but doesn't respond to attempts to connect, it's an invalid hostname, etc), then we try by name.

Basically, the two options in pseudocode:

1. `connect_to_cc(value)`
2. `connect_to_cc(get_cc_by_name(value))`

Does that make more sense?

catt/cli.py
skorokithakis: Oh, with `value`? That's fine, as long as the docstring explains what the value can be.

theychx:
> Does that make more sense?

Like this?
```
    if device:
        cast = get_chromecast_with_ip(device, DEFAULT_PORT)
    if not cast:
        if device and is_ipaddress(device):
            cast = get_chromecast_with_ip(device, DEFAULT_PORT)
            if not cast:
                msg = "No device found at {}".format(device)
                raise CastError(msg)
        else:
```

Looks a bit weird. :smiley:

skorokithakis:
Did you mean the second one to be `get_chromecast_by_name`?

skorokithakis:
I meant something more like:

`cast = get_chromecast_with_ip(device) or get_chromecast_by_name(device)`

As long as it's not twice as slow, because I don't want to make catt slower for a rarely-used feature. If it makes it any slower, your current PR is fine.

theychx:
Ah yes sorry, this is intentional, as I couldn't figure out how to implement the way you suggested.

We try to connect directly, treating `device` as a hostname. If that doesn't work, we proceed to the logic I already implemented.
Did I miss something?

theychx:
And my suggestion only introduces minimal additional latency.

skorokithakis:
Ah now I'm a bit confused :( In the example in your comment, I see it trying to connect by IP, if that fails, it checks whether the value is an IP address and tries to connect again. Why does it try to connect twice in exactly the same way?

theychx:
Yeah it's crap. :smile:  I think you meant:
```
if device:
    cast = get_chromecast_with_ip(device, DEFAULT_PORT)
if not cast:
    cache = Cache()
    cc_ip, cc_port = cache.get_data(device)

    if cc_ip:
        cast = get_chromecast_with_ip(cc_ip, cc_port)
    if not cast:
        cast = get_chromecast(device)
        if not cast:
            msg = 'Specified device "{}" not found'.format(device) if device else "No devices found"
            raise CastError(msg)
```

skorokithakis:
Ah, yep, that's what I meant! Anyway, I'm good with your current solution as well, so I will approve!

PR #205, opened 2019-08-11
Change the app name on the YouTube controller

skorokithakis:
Like this says, I didn't like the text "Python" on the YouTube display app, so I hackishly changed it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,8 +4,8 @@ repos:
     hooks:
     - id: black
       args: [--line-length=120]
--   repo: git://github.com/doublify/pre-commit-isort
-    rev: v4.3.0
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v4.3.21
     hooks:
     - id: isort
 -   repo: https://github.com/pre-commit/pre-commit-hooks
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Features
+
+* Add ability to use ip-address as device argument  (#197) [theychx]
+
 ### Fixes
 
 * Try to eliminate spurious error msgs when seeking (#194) [theychx]
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -1,3 +1,4 @@
+# isort:imports-stdlib
 import hashlib
 import json
 import tempfile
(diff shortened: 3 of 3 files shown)


skorokithakis:
I am currently trying to get isort to leave my shit alone.

skorokithakis:
The fucked up thing is that isort works locally, but not in precommit, even though they are the same version.

skorokithakis:
> And besides, wouldn't it be easier to just submit a PR to the casttube repo? I'm sure he wouldn't mind changing that screen text.

Change it to what, though? It's not going to be a parameter that goes through pychromecast.

What's your specific criticism of this?

theychx:
> Change it to what, though? It's not going to be a parameter that goes through pychromecast.

Yeah, fair point.
> What's your specific criticism of this?

I only have a very broad criticism of this, and is that is that it is extremely hackish. Why would we do something like this, to fix such a minor problem?

skorokithakis:
Well, for one, it's not a minor problem to me, I think it looks sloppy to have the app show "Python", of all things. Secondly, there isn't really a downside to this. Because of the try block, any error that happens causes the entire block to just get ignored, so there's no downside.

skorokithakis:
Bump

theychx:
I agree with you that it is suboptimal that the screen text says "Python", it should of course say "Catt" or "Cast All The Things" or whatever. The methology of this PR just bothers me, and the super-arrogant comment you put in the code doesn't help either.

I'm sure it is possible for `casttube` to implement the ability to have the text be whatever we like.

skorokithakis:
Even if `casttube` implements the ability, `pychromecast` still needs to expose it.

What's arrogant about the comment?

theychx:
> Even if casttube implements the ability, pychromecast still needs to expose it 

Yes, but only the included YouTube controller, which is effectively maintained by `casttube`s author.
> What's arrogant about the comment?

*"Some might say this goes too far. I say it doesn't go too far enough"*

Thats pretty arrogant.

skorokithakis:
> Yes, but only the included YouTube controller, which is effectively maintained by casttubes author.

Sure, but we don't use it directly, so how would we pass our text to it?

> Thats pretty arrogant.

https://www.youtube.com/watch?v=UUMdCV-Z7kk

theychx:
> Sure, but we don't use it directly, so how would we pass our text to it?

Eh? 
https://github.com/skorokithakis/catt/blob/eb3e20a8d2629799ff0e0075746f1fa7f2c4f525/catt/controllers.py#L15
https://github.com/skorokithakis/catt/blob/eb3e20a8d2629799ff0e0075746f1fa7f2c4f525/catt/controllers.py#L605
> https://www.youtube.com/watch?v=UUMdCV-Z7kk

LOL, if your intention was to reference that clip, then it is not arrogant, obviously. Pretty easy to miss that context though.

skorokithakis:
>  self._controller = YouTubeController() 

Exactly, so the author of PyChromecast would have to add a parameter too, since we don't instantiate YouTubeSession ourselves.

> Pretty easy to miss that context though.

True, it's an old show. I thought the "too far enough" would have tipped you off though :P

theychx:
> Exactly, so the author of PyChromecast would have to add a parameter too, since we don't instantiate YouTubeSession ourselves.

Yes, the `casttube` author would have to do a PR for `pychromecast`, to update the controller. Which he would have to do anyway, in the case of a new `casttube` release with new functionality.

theychx:
> I thought the "too far enough" would have tipped you off though :P

Hehe, yeah, I've always been kind of slow in the uptake. :smile:

skorokithakis:
Yeah, so we'd have to wait for two pieces of software to be updated. Anyway, if you don't think it's worth it, I'm good with that.

PR #207, opened 2019-08-16
feat: Add ability to use subtitles with remote content

### LINKED ISSUE #199: Allow specifying subtitles when casting remote url
Currently subtitle_url is only set here for local files:
https://github.com/skorokithakis/catt/blob/master/catt/cli.py#L186

It should be possible to specify a remote subtitle to go with a remote file eg:

catt cast -s "http://127.0.0.1/subs.vtt" "http://127.0.0.1/file.mp4"

theychx:
fixes #199

This turned out to be a bit more work than I anticipated, but the previous approach just didn't cut it anymore.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -113,6 +113,13 @@ def write_config(settings):
         raise CliError("No device specified")
 
 
+def create_server_thread(server_args):
+    thr = Thread(target=serve_file, args=server_args)
+    thr.setDaemon(True)
+    thr.start()
+    return thr
+
+
 def process_subtitles(ctx, param, value):
     if not value:
         return None
@@ -157,9 +164,7 @@ def cast(settings, video_url, subtitles, force_default, random_play, no_subs, no
     )
 
     if stream.is_local_file:
-        st_thr = Thread(target=serve_file, args=(video_url, stream.local_ip, stream.port, stream.guessed_content_type))
-        st_thr.setDaemon(True)
-        st_thr.start()
+        st_thr = create_server_thread((video_url, stream.local_ip, stream.port, stream.guessed_content_type))
     elif stream.is_playlist and not (no_playlist and stream.video_id):
         if stream.playlist_length == 0:
             cst.kill(idle_only=True)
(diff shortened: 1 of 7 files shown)


catt/cli.py
```diff
@@ -206,23 +179,32 @@ def cast(settings, video_url, subtitle, force_default, random_play, no_subs, no_
         video_id = stream.video_id or stream.playlist_all_ids[0]
         cst.play_playlist(stream.playlist_id, video_id=video_id)
     else:
+        if not subtitles and not no_subs and stream.is_local_file:
+            subtitles = hunt_subtitles(video_url)
+        if subtitles:
+            subs = SubsInfo(subtitles, stream.local_ip, stream.port + 1)
+            su_thr = Thread(target=serve_file, args=(subs.file, stream.local_ip, subs.port, "text/vtt;charset=utf-8"))
```
skorokithakis: We might want to abstract these three lines into a function, so we don't repeat them every time.

catt/subs_info.py
```diff
@@ -0,0 +1,51 @@
+import re
+
+import requests
+
+from .error import SubtitlesError
+from .util import create_temp_file
+
+
+class SubsInfo:
```
skorokithakis: We probably need docstrings here, It's not entirely clear at a glance what this class does.

catt/subs_info.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, subs_url: str, local_ip: str, port: int) -> None:
+        self._subs_url = subs_url
+        self._local_ip = local_ip
+        self.port = port
```
skorokithakis: Why is `port` not `_port`? (Add a comment explaining, if it's intentional).

catt/subs_info.py
```diff
@@ -0,0 +1,51 @@
...
+
+        ext = subs_url.lower().split(".")[-1]
+        if "://" in subs_url:
```
skorokithakis: I think this could be a bit more elegant like so:

```
# Read subs (local or remote) into memory.
subs = read_subs(subs_url)

# If the file is in the SRT format, convert to WebVTT.
if ext == "srt":
    subs = self._convert_srt_to_webvtt(subs)

# Write the temp file with the subs.
self.file = create_temp_file(subs)
```

We're always writing the temp file in this case, but I think that's worth it to simplify the code. `read_subs` will either fetch the remote subs or read local subs, basically `_fetch_remote_subs` or `_read_srt_subs` depending on what's going on in the URL.

skorokithakis:
This is great! I've added a few comments.

catt/subs_info.py
theychx: I was just thinking that making `self._local_ip` public would be superflous, as the local ip is already exposed as StreamInfo.local_ip. But yeah, you're right, it looks strange, so maybe it should just be:
```
self.local_ip = local_ip
self.port = port
```

I agree with all your other comments, and will apply fixes.

skorokithakis: Ah, I see. Either way is good with me

theychx:
Done!

skorokithakis:
Looks great!

PR #208, opened 2019-08-17
fix: Adapt write_config to "new" get_chromecast

theychx:
:face_with_head_bandage: I completely missed this one.
Ever since I did the api work, `get_chromecast` no longer throws anything, if supplied with name of non-existent cc, which we have relied on in `write_config`. But fixed now...


skorokithakis:
Oops!

PR #209, opened 2019-08-18
Late summer cleaning

theychx:
:smiley:


catt/cli.py
```diff
@@ -88,6 +99,13 @@ def process_device(ctx, param, value):
         return ctx.default_map["aliases"].get(value, value)
 
 
+def create_server_thread(server_args):
+    thr = Thread(target=serve_file, args=server_args)
```
skorokithakis: DIdn't we just abstract this in another function in the last PR?

theychx: It's the same one. I just moved it.

skorokithakis: Ah, right.

PR #213, opened 2019-10-07
Local ip fix

### LINKED ISSUE #212: TypeError: Str, bytes or bytearray expected, not NoneType
Great project. 
I plan to use it with a script to play customised MP4 videos to visitors when they open the door with their access card.

I am using catt on Win10pro x64, using catt v0.10.0 and python3.7.4. 
The Chromecast device is Gen2, located on a separate VLAN
I connect to it directly over IP (no proxy) and it opens connections on TCP8008 and TCP8009
I can use catt to stream utube video without any issue.

However when I tried MP4 files it fails. See output in attached error.txt file
[error.txt](https://github.com/skorokithakis/catt/files/3694148/error.txt)

I initially thought I was using the wrong syntax to launch the program but https://www.computerhope.com/issues/ch001980.htm seems to use the same syntax without issue, albeit to a local chromecast.

I am not a coder but the error suggests it is expecting a value that is not there.

So I began to suspect there was corruption in the file itself which was MOV format converted by VLC to MP4. So I tried a sample from "https://file-examples.com/index.php/sample-video-files/sample-mp4-files" and got the same error.

Finally I downloaded the file of the utube video that I knew I could stream OK. But alas the same error.

Can you suggest how to resolve the issue. Thx

theychx:
Would something like this be acceptable as a solution to #212?

In case you are wondering; I was originally keen on scrapping `netifaces` as a dependency, because it, unlike our other dependencies, needs to be compiled.

fixes #212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## Unreleased
+
+### Fixes
+
+* Only serve files forever if it is necessary (#210) [Stavros Korokithakis]
+
+
 ## v0.10.0 (2019-08-19)
 
 ### Features
--- a/catt/util.py
+++ b/catt/util.py
@@ -1,11 +1,16 @@
 import ipaddress
 import json
+import sys
 import tempfile
 import time
 from pathlib import Path
 
 import click
-import ifaddr
+
+try:
+    import netifaces
+except ImportError:
+    import ifaddr
 
 
 def warning(msg):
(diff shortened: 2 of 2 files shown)


skorokithakis:
Hmm, what does `netifaces` do that we can't do with `get_adapters()`? I think that piece of code needs some comments detailing what it's doing at each step and why, it is currently a bit dense.

theychx:
> Hmm, what does netifaces do that we can't do with get_adapters()?

`netifaces.gateways()` provides a "default" entry, which changes dynamically when f.ex. you plug in your box to a wired network. `ifaddr.get_adapters()` provides no such thing, so currently, we are guessing the local ip from the cc ip.
> I think that piece of code needs some comments detailing what it's doing at each step and why, it is currently a bit dense

I agree, I just wanted to float the idea, hence the draft status of this PR. We should also throw a real error msg to the cli user and update the readme, so users with advanced network setups can install `netifaces` manually if needed.

skorokithakis:
It looks good to me, but I'm not very eager to reintroduce `netifaces` as a dependency... Is there a way we can get the default entry ourselves perhaps?

theychx:
Neither am I, but as I mentioned in the ticket comments, I have searched in vain for alternative solutions to this problem. We could of course also just leave it be, and tell people that the box running `catt` and the cc need to be on the same subnet. Which would be fine with me.

skorokithakis:
An easier middle ground would be to give them a cli switch to specify the address to listen on themselves, I guess. I think that's better than reintroducing `netifaces` or doing nothing (and shouldn't be too hard, no?).

theychx:
Yeah, you're right, that's probably the least ugly solution. I'll close this PR then, and implement the cli switch.

PR #216, opened 2019-10-15
fix: Add socket based fallback approach in get_local_ip

### LINKED ISSUE #212: TypeError: Str, bytes or bytearray expected, not NoneType

theychx:
I found an alternative `socket` based approach to determinig local ip, that should work in cases such as #212.
The local ip cli option turned out to not be as trivial to implement as I had anticipated, so I cooked up this, which seems like a better solution.

fixes #212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixes
 
+* Add socket based fallback approach in get_local_ip. [theychx]
+
 * Fix namespace related crash when using DashCast. [Stavros Korokithakis]
 
 * Only serve files forever if it is necessary (#210) [Stavros Korokithakis]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixes
 
+* Fix namespace related crash when using DashCast. [Stavros Korokithakis]
+
 * Only serve files forever if it is necessary (#210) [Stavros Korokithakis]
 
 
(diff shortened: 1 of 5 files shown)


theychx:
Waiting for @drareve to test this...
(It works on my box/setup)

theychx:
Ready for review.

skorokithakis:
I think the changelog software messes up and attributes things to me because I'm the one who merged, I will leave merges to you from now on so you get proper attribution.

PR #217, opened 2019-10-17
Cache zeroconf device info

theychx:
Howdy
For some time now, Google devices have stopped providing manufacturer/model_name info through DIAL, breaking our device type detection stuff in `StreamInfo`. So I've implemented a change to our caching, so we cache all the info we need from zeroconf. This work also turned out make some code a bit leaner.

PS: Your changelog thingy is indeed out of control :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixes
 
+* Add socket based fallback approach in get_local_ip. [Stavros Korokithakis]
+
 * Add socket based fallback approach in get_local_ip. [theychx]
 
 * Fix namespace related crash when using DashCast. [Stavros Korokithakis]
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -20,7 +20,6 @@ from .stream_info import StreamInfo
 from .util import is_ipaddress, warning
 
 NO_PLAYER_STATE_IDS = [DASHCAST_APP_ID]
-DEVICES_WITH_TWO_MODEL_NAMES = {"Eureka Dongle": "Chromecast"}
 DEFAULT_PORT = 8009
 VALID_STATE_EVENTS = ["UNKNOWN", "IDLE", "BUFFERING", "PLAYING", "PAUSED"]
 
@@ -75,8 +74,9 @@ def get_cast(device=None):
 
     :param device: Can be an ip-address or a name.
     :type device: str
-    :returns: Chromecast object for use in a CastController.
-    :rtype: pychromecast.Chromecast
+    :returns: Chromecast object for use in a CastController,
+              and CCInfo object for use in setup_cast and StreamInfo
+    :rtype: (pychromecast.Chromecast, CCInfo)
     """
 
     cast = None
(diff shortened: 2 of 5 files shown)


catt/controllers.py
```diff
@@ -215,32 +219,33 @@ def __init__(self):
...
+            return None
         if name:
             fetched = data.get(name)
         else:
             # When the user does not specify a device, we need to make an attempt
             # to consistently return the same IP, thus the alphabetical sorting.
             fetched = data[min(data, key=str)]
-        return (fetched["ip"], fetched.get("group_port", 0)) if fetched else (None, None)
+        return (
+            CCInfo(fetched["ip"], fetched["port"], fetched["manufacturer"], fetched["model_name"], fetched["cast_type"])
```
theychx: Is there a more elegant way of doing this?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 ### Fixes
 
+* Add socket based fallback approach in get_local_ip. [Stavros Korokithakis]
```
skorokithakis: Oh geez.

catt/stream_info.py
```diff
@@ -7,27 +7,27 @@
 from .util import get_local_ip, guess_mime
 
 AUDIO_DEVICE_TYPES = ["audio", "group"]
-ULTRA_MODELS = [("Xiaomi", "MIBOX3"), ("Google Inc.", "Chromecast Ultra")]
+ULTRA_MODELS = [("Xiaomi", "MIBOX3"), ("Unknown manufacturer", "Chromecast Ultra")]
```
theychx: The **manufacturer** field is only available though DIAL, so this is now completely unavailable for Google devices.

catt/controllers.py
skorokithakis: Yep, `CCInfo(**fetched)`.

theychx:
@skorokithakis 
Just FYI, it turns out that those missing info fields are indeed still available via DIAL. It's just not where `pychromecast` is currently looking.

The info is still available here: `http://CAST_IP:8008/ssdp/device-desc.xml`
`pychromecast` is currently using `http://CAST_IP:8008/setup/eureka_info?options=detail`

The problem with the former is that some devices (my cc audio included) respond with a 404, so the `pychromecast` folks actually [switched](https://github.com/balloob/pychromecast/pull/197) from the former to the latter some time ago.

So in conclusion, I still think we should go ahead with this PR.

skorokithakis:
That makes sense, although I wonder why they didn't just fall back to the latter of the former failed. Do you want to address the keyword arguments line first or should I approve now?

catt/controllers.py
theychx: How can I be sure that the params are fed to `CCInfo` in the right order?

theychx:
> although I wonder why they didn't just fall back to the latter of the former failed

At the time of the PR, it seems like most devices 404'ed on `device-desc.xml`. And besides, `eureka_info` did at that point contain all the info needed.

catt/controllers.py
skorokithakis: They're passed as keyword parameters, so order doesn't matter.

theychx: Cheers. The relation between positional and keyword parameters is confusing in python. Any best practice you can recommend?

skorokithakis: Bah I pressed resolve on this by accident, the mobile interface is confusing. Hmm, I tend to call things positionally for small numbers of arguments, if the number of arguments gets larger or they're more convoluted, I tend to specify keywords.

In this case, it's just easier to expand into keywords, rather than specify the parameters one by one.

theychx:
@skorokithakis Did you have any further comments?

catt/controllers.py
```diff
@@ -215,32 +219,29 @@ def __init__(self):
...
             fetched = data.get(name)
         else:
             # When the user does not specify a device, we need to make an attempt
             # to consistently return the same IP, thus the alphabetical sorting.
             fetched = data[min(data, key=str)]
-        return (fetched["ip"], fetched.get("group_port", 0)) if fetched else (None, None)
+        return CCInfo(**fetched) if fetched else None
 
-    def set_data(self, name: str, ip: str, port: int) -> None:  # type: ignore
+    def set_data(self, name: str, device_entry: CCInfo) -> None:  # type: ignore
```
skorokithakis: What's the error here that made us ignore the type?

catt/stream_info.py
skorokithakis: Should this be a comment in the code?

theychx: It was FYI, but a comment won't hurt. :smiley:

catt/controllers.py
theychx: Removing the ignore results in this error msg from `mypy`:
```
Signature of "set_data" incompatible with supertype "CattStore"
```
..which still doesn't make sense to me.

PR #219, opened 2019-10-21
Refactor of namespace fix

theychx:
..and a little test. :desktop_computer:

### DIFF
--- a/realcc_tests/test_procedure.py
+++ b/realcc_tests/test_procedure.py
@@ -187,7 +187,8 @@ DASHCAST_CTRL_TESTS = [
     CattTest(
         "cast GitHub website frontpage",
         ["cast_site", "https://github.com"],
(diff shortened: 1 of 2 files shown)


realcc_tests/test_procedure.py
```diff
@@ -183,6 +183,14 @@ def run(self) -> bool:
     ),
 ]
 
+DASHCAST_CTRL_TESTS = [
+    CattTest(
+        "cast GitHub website frontpage",
+        ["cast_site", "https://github.com"],
+        check_data=("status_text", "The world’s leading software development platform · GitHub"),
```
skorokithakis: This might be a bit too specific, can we check that only `GitHub` is in the data? We don't want tests breaking if they change anything about their tagline...

theychx: Fixed.

skorokithakis:
Looks good, thanks!

PR #222, opened 2019-11-19
Clarify installation instructions

theychx:
:smiley:


skorokithakis:
Hmm, this is a bit controversial because in systems that don't have Python 2 installed, pip points to 3 anyway. However, I guess there always exists the pip3 executable, so I guess it's fine.

PR #229, opened 2020-01-03
Mixcloud fix

theychx:
(and an update for the gitignore) :tv:


theychx:
Yeah, I snatched it from https://github.com/github/gitignore. To lazy to trim it. :smile:

PR #237, opened 2020-02-12
feat: Add block option to cast cmd

### LINKED ISSUE #218: Catt does not exit after playback ends
As the title states, catt simply does not exit after playback is complete. In catt-qt, I run catt instead of using catt api to play, because catt can play files and catt api can't. It runs catt in the background and this works with the exception of leaving running catt processes behind. I suppose it could try to kill catt processes on idle state but this seems sketchy at best. Is there a reason why catt doesn't exit after playback ends?

theychx:
As promised.

Fixes #218

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -117,7 +117,7 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 @click.group(context_settings=CONTEXT_SETTINGS)
 @click.option("--delete-cache", is_flag=True, help="Empty the Chromecast discovery cache.")
 @click.option("-d", "--device", metavar="NAME_OR_IP", callback=process_device, help="Select Chromecast device.")
-@click.version_option(version=__version__, prog_name="catt", message="%(prog)s v%(version)s, Xylophone Xtravaganza.")
+@click.version_option(version=__version__, prog_name="catt", message="%(prog)s v%(version)s, Yearning Yachtman.")
 @click.pass_context
 def cli(ctx, delete_cache, device):
     if delete_cache:
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -163,8 +163,9 @@ def write_config(settings):
     "-b",
     "--block",
     is_flag=True,
-    help="Keep catt process alive until playback has ended "
-    "(currently exits after playback of single media, so not useful with playlists yet).",
+    help="Keep catt process alive until playback has ended. "
+    "Only useful when casting remote files, as catt is already running a server when casting local files. "
+    "Currently exits after playback of single media, so not useful with playlists yet.",
 )
 @click.pass_obj
 def cast(settings, video_url, subtitles, force_default, random_play, no_subs, no_playlist, ytdl_option, block):
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -159,8 +159,15 @@ def write_config(settings):
     help="YouTube-DL option. "
     "Should be passed as `-y option=value`, and can be specified multiple times (implies --force-default).",
 )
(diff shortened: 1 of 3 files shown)


theychx:
@skorokithakis Ready for review. The tests are failing because of random youtube whining about too many requests.
And I screwed up your changelog. Sorry about that.

catt/cli.py
```diff
@@ -159,15 +159,24 @@ def write_config(settings):
...
+    "-b",
+    "--block",
+    is_flag=True,
+    help="Keep catt process alive until playback has ended. "
+    "Only useful when casting remote files, as catt is already running a server when casting local files. "
+    "Currently exits after playback of single media, so not useful with playlists yet.",
+)
 @click.pass_obj
-def cast(settings, video_url, subtitles, force_default, random_play, no_subs, no_playlist, ytdl_option):
+def cast(settings, video_url, subtitles, force_default, random_play, no_subs, no_playlist, ytdl_option, block):
```
skorokithakis: Perhaps we should give this new option a default value of `False`. I wouldn't want to break any existing callers if we can help it.

realcc_tests/test_procedure.py
```diff
@@ -208,8 +208,8 @@ def run(self) -> bool:
     ),
     CattTest(
         "cast first video from youtube playlist on default controller",
-        ["cast", "https://www.youtube.com/watch?v=J4R7M4xH_BA&list=PLhReEC3095X2aQ29-iW2bto0cYPaxch75"],
-        check_data=("status_text", "Casting: The Warrior: A 2015 Psychedelic Trance Mix"),
+        ["cast", "https://www.youtube.com/watch?v=jSL1nXza7pM&list=PLAxEbmfNXWuIhN2ppUdbvXCKwalXYvs8V&index=2&t=0s"],
```
skorokithakis: Is this going to work with YouTube if we rerun the tests? I wouldn't want our tests to always just fail from now on.

skorokithakis:
Thanks! I've left some comments.

skorokithakis:
Also, Yearning Yachtman.

catt/cli.py
theychx: `is_flag` is set in the `click` decorator, so it already defaults to `False`.

realcc_tests/test_procedure.py
theychx: Sorry, I meant in the regular test-suite, which makes CI fail. But I can see that I forgot to remove some params from the url. Maybe we'll be lucky after a new commit. :smiley:

catt/cli.py
skorokithakis: For us, but if a third-party script is calling `cast` directly, it'll fail. I agree that it's not public and we have an API and people shouldn't be calling it, but I'd rather potentially save a few people the hassle since we can just add a `= False`.

theychx: Fair enough, I'll fix it.

realcc_tests/test_procedure.py
theychx: Just to be clear, the above test in `test_procedure` succeeds.

skorokithakis: Right, I just marked this as a generic way to leave a comment. I'm not quite clear which one fails, I just don't want a flappy test suite :P

theychx:
:tada: All hail the mercyful YouTube gods.

skorokithakis:
Can you squash and merge? I don't want to take away your credit.

theychx:
> Yearning Yachtman is nearing completion

Indeed. :smiley: I've got a couple more PR's coming up.

PR #238, opened 2020-02-17
feat: Add force option to stop cmd

### LINKED ISSUE #236: Casting not stopped with Smart Display/Speaker when using local media
There are 2 symptoms to this issue, however my hunch is that they're both the same cause.  This happens on both my Google Nest Mini, as well as my Google Nest Hubs.  To test, I have the cast initiation sound switched on

### **Cast session stays connected after locally played media ends.**
**What Happens**
Casting a local file to my Nest Hub plays the file, however the cast session stays connected after the media finishes playing as per the below image.  I can tell this as the Google Nest Hub displays the cast logo as below, and for both the speaker and hub, when another file is cast to the devices, the cast initiation sound does not play.

```catt -d "Office Hub" cast C:\Users\Greg\Documents\Projects\assistant-relay\relay\bin\sounds\doorbell.mp3```

image

Attempting a stop command does not kill the session either ```catt -d "Office Hub" stop``` and leaves the Smart Display on the above screen.

Eventually, the cast session times out and the Smart Display resumes its normal operation

I have also tried this with an MP4 file, and can confirm that whilst it is playing, the stop command does not stop the cast.

theychx:
..as a workaround for devices not responding to stop cmd under certain circumstances.

fixes #236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Features
 
-* Add block option to cast cmd. [theychx]
+* Add block option to cast cmd (#237) [theychx]
 
 
 ## v0.10.3 (2020-01-26)
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -300,7 +300,7 @@ def play(settings):
     "-f",
     "--force",
     is_flag=True,
-    help="Start dummy chromecast app before sending stop command "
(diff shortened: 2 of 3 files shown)


skorokithakis:
Sorry this took so long, looks good!

theychx:
No worries!

PR #239, opened 2020-02-18
feat: Add play_toggle cmd

theychx:
:tv: :cowboy_hat_face: :notes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


theychx:
This youtube whining is getting on my nerves. Is there a way to trigger a new CI build without pushing a new commit?

skorokithakis:
Approven, though flappy tests sadden me. We should also add Python 3.7 to our test suite at some point.

PR #243, opened 2020-02-26
Make new config cmds work with ip as device arg

theychx:
:computer:

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -11,7 +11,7 @@ from urllib.parse import urlparse
 import click
 
 from . import __version__
-from .controllers import Cache, CastState, StateFileError, StateMode, get_chromecast, get_chromecasts, setup_cast
+from .controllers import Cache, CastState, StateFileError, StateMode, get_chromecasts, setup_cast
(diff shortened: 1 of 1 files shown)


catt/cli.py
```diff
@@ -517,7 +517,13 @@ def get_alias_from_config(config, device):
 def get_device_from_settings(settings):
     device = settings.get("device")
     if device:
-        if not get_chromecast(device):
+        devices = get_chromecasts()
+        try:
+            if is_ipaddress(device):
+                next(d for d in devices if d.host == device)
+            else:
+                next(d for d in devices if d.name == device)
```
skorokithakis: Can you refactor these in a function called `first()` or similar? It'll be clearer what it does.

theychx: How would I change the specific attribute I need to check in the if clause?

theychx: Or did you just mean put this part of the logic in its own func?

skorokithakis: Sorry, I just noticed that the attribute was different. Never mind then.

PR #267, opened 2020-06-08
Update requirements to latest versions

Poorchop:
Update youtube-dl, PyChromecast, click, ifaddr, and requests to the latest versions, fixing broken casting of Twitch.tv

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,11 +21,11 @@ with open("README.rst") as readme_file:
     readme = readme_file.read()
 
 requirements = [
-    "youtube-dl>=2019.01.24",
-    "PyChromecast>=4.1.1",
-    "Click>=5.0",
-    "ifaddr>=0.1.4",
-    "requests>=2.18.4",
+    "youtube-dl>=2020.6.6",
+    "PyChromecast>=6.0.0",
+    "Click>=7.1.2",
+    "ifaddr>=0.1.7",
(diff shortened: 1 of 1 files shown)


skorokithakis:
This is good, thank you. However, doesn't it already work if you reinstall catt, since it's >=?

Poorchop:
I'm pretty sure that I tried to reinstall catt but it still wasn't working for me. I remember it saying that I had the latest version of everything. Twitch wasn't working for me though and I finally realized that it was because it was using an old version of youtube-dl and reinstalling catt wasn't fetching the latest version.

Poorchop:
That was the plan but I was more concerned about catt being broken for future users. It only occurred to me after this discussion that new installs would be safe since they would pull the latest version of youtube-dl and only existing installs are at the risk of breaking.

This also makes me wonder if it would be a good idea to check if the user already has youtube-dl installed prior to pulling it in as a requirement. I already have it installed via my package managers in Linux and OS X and in Windows, I download the binary and manually update it as needed. Maybe just checking if it is in the user's PATH first would be good to avoid multiple installs/managing updates?

skorokithakis:
That would be useful but I don't want to get into messing with the user's environment without confirmation, I'd rather they did it themselves. We could detect that YouTube-dl didn't work and print a message to upgrade it, though.

Issue #271, opened 2020-06-29
Pychromecasts last release breaks casting


sickmartian:
Seems like since the last release get_chromecasts will return a tuple: https://github.com/home-assistant-libs/pychromecast/commit/e8d04cd9a12b4f434f69a98f2eb03d44fbc302ae

Breaking change: https://github.com/home-assistant-libs/pychromecast/blob/e8d04cd9a12b4f434f69a98f2eb03d44fbc302ae/pychromecast/__init__.py#L161

Might be good to cap the version of the library to 6.x

It generates the following error for me:
```Traceback (most recent call last):
  File "/usr/local/bin/catt", line 11, in <module>
    sys.exit(main())
  File "/home/osmc/.local/lib/python3.6/site-packages/catt/cli.py", line 570, in main
    return cli(obj={}, default_map=get_default_map())
  File "/home/osmc/.local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/osmc/.local/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/osmc/.local/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/osmc/.local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/osmc/.local/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/osmc/.local/lib/python3.6/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/osmc/.local/lib/python3.6/site-packages/catt/cli.py", line 165, in cast
    settings["device"], video_url=video_url, prep="app", controller=controller, ytdl_options=ytdl_option
  File "/home/osmc/.local/lib/python3.6/site-packages/catt/controllers.py", line 144, in setup_cast
    cast, cc_info = get_cast(device_name)
  File "/home/osmc/.local/lib/python3.6/site-packages/catt/controllers.py", line 92, in get_cast
    cache = Cache()
  File "/home/osmc/.local/lib/python3.6/site-packages/catt/controllers.py", line 222, in __init__
    devices = get_chromecasts()
[...]

PR #272, opened 2020-06-29
fix: Force pychromecast 7, update to new api

IanCal:
Pychromecast 7 has a breaking change, with the return type of the
get_chromecasts function changing to also return a browser.

I ran black over this file but for me it changes a lot more of the formatting - happy to do that but wanted to check (and also at least isolate the actual change made to make it clearer). Also running tests locally seemed to be trying to do things to my actual chromecasts (?) so I thought I'd leave that for travis.

Deals with issue #271

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -41,7 +41,7 @@ APPS = [
 
 
 def get_chromecasts():
-    devices = pychromecast.get_chromecasts()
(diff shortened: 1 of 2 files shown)


skorokithakis:
This looks good, thank you! You should run `pre-commit`, not just `black`, as `pre-commit` has some additional settings. It doesn't look like CI will fail, though.

PR #275, opened 2020-07-01
Pychromecast7

theychx:
Ongoing......... :tv:

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -13,6 +13,7 @@ from pychromecast.config import APP_MEDIA_RECEIVER as MEDIA_RECEIVER_APP_ID
 from pychromecast.config import APP_YOUTUBE as YOUTUBE_APP_ID
 from pychromecast.controllers.dashcast import DashCastController as PyChromecastDashCastController
 from pychromecast.controllers.youtube import YouTubeController
+from pychromecast.discovery import get_host_from_service_info, get_info_from_service
 
 from . import __version__
 from .error import AppSelectionError, CastError, ControllerError, ListenerError, StateFileError
@@ -23,6 +24,7 @@ GOOGLE_MEDIA_NAMESPACE = "urn:x-cast:com.google.cast.media"
 DEFAULT_PORT = 8009
 VALID_STATE_EVENTS = ["UNKNOWN", "IDLE", "BUFFERING", "PLAYING", "PAUSED"]
 CLOUD_APP_ID = "38579375"
+GOOGLE_ZCONF = "_googlecast._tcp.local."
 
 
 class App:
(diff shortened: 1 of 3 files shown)


skorokithakis:
Thank you for this. While you're developing, can you add function types where practical? Currently we don't indicate what we accept or return, and it's making development hard. I tried to add some but discovery isn't working for me and couldn't figure out what functions were supposed to return.

catt/controllers.py
```diff
@@ -41,23 +43,43 @@ def __init__(self, app_name, app_id, supported_device_types):
...
+        devices, zconf = get_chromecasts()
+    return [get_ccinfo(d, zconf) for d in devices]
+
+
+def get_ccinfo(device, zconf=None):
+    if zconf is None:
+        ip, port = device.host, device.port
+    else:
+        service_name = (
+            "-".join(device.model_name.split()) + "-" + str(device.uuid).replace("-", "") + "." + GOOGLE_ZCONF
```
theychx, line 62: As per [this](https://github.com/home-assistant-libs/pychromecast/issues/381#issuecomment-651945670) comment, the `service_name` should be available from the `Chromecast` object, but I'm struggling to find just that.

skorokithakis:
To clarify, I'm not saying add types to everything, just whatever function you happen to touch and is easy.

catt/controllers.py
skorokithakis: I'll give it a shot, I'll reconfigure my network so I can do discovery and look for it.

theychx:
Gotcha. Basic casting should be working. I just need to get the auxilliary stuff up to date.

skorokithakis:
I resolved some conflicts and will look for the `service_name` now.

theychx:
There's a new `pychromecast` PR: https://github.com/home-assistant-libs/pychromecast/pull/385
I suggest we wait and see how that pans out, before putting more effort into this.

skorokithakis:
Ah, agreed. I will do some work adding types to master in the mean time. What do you think regarding using PyChromecast's socket to talk to the Chromecast? They seemed to indicate that the socket is better for when groups change their leader, etc.

theychx:
> What do you think regarding using PyChromecast's socket to talk to the Chromecast

That shouldn't be neccesary. Our current crude approach in `get_cast` already handles this (i.e. connection to the old leader fails, we do discovery and find the new leader). If anything, we should use the new `get_listed_chromecasts` to do "re-discovery" in these cases.

skorokithakis:
@theychx the PR was merged, which is certainly interesting. Let's hope they keep that method of discovery and don't break it later...

PR #277, opened 2020-07-05
Pychromecast 7.5 compat

theychx:
..with no cache! (only marginally slower than before)

Let me know what you think.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -3,6 +3,7 @@ import threading
 from enum import Enum
 from pathlib import Path
 from typing import Any
+from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Union
@@ -76,7 +77,7 @@ def get_cast_devices(names: Optional[List[str]] = None) -> List[CastDevice]:
     return devices
 
 
-def get_cast_devices_info():
+def get_cast_devices_info() -> Dict[str, Dict[str, Union[str, int]]]:
     devices = get_cast_devices()
     return {d.cast.name: d.info for d in devices}
 
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -66,6 +66,16 @@ class CastDevice:
 
 
 def get_cast_devices(names: Optional[List[str]] = None) -> List[CastDevice]:
+    """
+    Discover all available devices, optionally filtering them with list of specific device names
+    (which will speedup discovery, as pychromecast does this in a non-blocking manner).
+
+    :param names: Optional list of device names.
+    :type names: List[str]
+    :returns: List of CastDevice wrapper objects containing cast object and additional ip/port info.
+    :rtype: List[CastDevice]
+    """
+
     if names:
         services, browser = pychromecast.discovery.discover_listed_chromecasts(friendly_names=names)
     else:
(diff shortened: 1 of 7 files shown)


skorokithakis:
Oooh very nice, thanks! I will review when I'm home and get back to you.

catt/controllers.py
```diff
@@ -61,24 +61,24 @@ def all_info(self):
...
+    if names:
+        services, browser = pychromecast.discovery.discover_listed_chromecasts(friendly_names=names)
+    else:
+        services, browser = pychromecast.discovery.discover_chromecasts()
+    pychromecast.stop_discovery(browser)
+    return [(pychromecast.get_chromecast_from_service(s, browser.zc), s[4]) for s in services]
+
+
+def get_chromecasts(names: Optional[List[str]] = None) -> List[pychromecast.Chromecast]:
+    devices = [d for d, _ in get_chromecasts_and_ips(names=names)]
```
theychx, line 74: This approach is a bit quirky, but it enables us to keep the `Zeroconf` instance inside of `get_chromecasts_and_ips`. Which also means that we don't need to import `zeroconf` only to keep `mypy` happy.

skorokithakis: Looks good to me.

catt/controllers.py
```diff
@@ -61,24 +61,24 @@ def all_info(self):
 ]
 
 
-def get_chromecasts() -> List[pychromecast.Chromecast]:
-    devices = pychromecast.get_chromecasts()
+def get_chromecasts_and_ips(names: Optional[List[str]] = None) -> List[Tuple[pychromecast.Chromecast, str]]:
```
skorokithakis, line 64: Maybe we should move these functions to a file called `discovery.py`? They seem only necessary for discovery and it would be much easier to understand the code if they were all groupped together.

catt/controllers.py
```diff
@@ -61,24 +61,24 @@ def all_info(self):
...
+    if names:
+        services, browser = pychromecast.discovery.discover_listed_chromecasts(friendly_names=names)
+    else:
+        services, browser = pychromecast.discovery.discover_chromecasts()
+    pychromecast.stop_discovery(browser)
+    return [(pychromecast.get_chromecast_from_service(s, browser.zc), s[4]) for s in services]
+
+
+def get_chromecasts(names: Optional[List[str]] = None) -> List[pychromecast.Chromecast]:
```
skorokithakis, line 73: These functions need a line or two of docstring, probably. Most are fairly straightforward in what they do, but there are some considerations (e.g. speed of discovery, whether they connect to the Chromecast or not, what a CCInfo object is) that isn't currently clear.

setup.py
```diff
@@ -20,7 +20,7 @@
 
 requirements = [
     "youtube-dl>=2020.6.6",
-    "PyChromecast~=6.0",
+    "PyChromecast>=7.1.0",
```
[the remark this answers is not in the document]
theychx: I understand your reasoning, but pins can be annoying when you're not installing to a virtenv (I'm guessing most of our users). And `pychromecast` haven't really been that unstable.
Another approach could be making `pipx install` the recommended way of installing `catt`, in which case the pin wouldn't be a problem.

catt/controllers.py
theychx: Fair point. I'll make a new PR after this is merged.

setup.py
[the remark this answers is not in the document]
skorokithakis: You mean because they may have other applications that depend on higher versions? I guess that's reasonable.

skorokithakis:
Merge at will!

skorokithakis:
What remains to be done for this PR? I've recently noticed that Pychromecast 6 doesn't always discover my Chromecast, while 7 does, so it would be good if we could merge the Pychromecast 7 compatibility changes in and fix the problem.

skorokithakis:
Actually, it might be that Pychromecast 7 doesn't solve the problem either, let me test some more.

skorokithakis:
Yeah, that's just the browser object, not the Chromecasts. Odd that discovery is failing like that.

theychx:
@skorokithakis Please review (again :smiley: ).

I can't figure out why ci-tests fail on py3.7.

skorokithakis:
You are a beautiful man. I will review ASAP!

skorokithakis (approved):
Looks good to me!

PR #295, opened 2020-09-21
fix: Increase timeout for the Chromecast to connect to the HTTP server

### LINKED ISSUE #294: Timeout when chromecast connecting to HTTP server.
Hi,

Tried catt for the first time yesterday. Really great, thanks.

One problem I had when serving a file from my laptop over wifi: the chromecast takes a little while to connect to the http server and this can cause issues.

I saw two failure modes:
 - catt would exit before the chromecast could connect.
 - the chromecast would connect and start playing the file, then catt would stop the http server.

The latter is presumably a race condition?

My guess is that catt waits for a fixed time (e.g. 5 seconds) for the chromecast to connect? I wonder if it just needs to wait a little longer? If that's the case, should the timeout be configurable?

I was able to get a video playing after several attempts though :)

Thanks

vext01:
The current 10 second wait isn't long enough on slow networks (like my wifi network at home!). Let's wait 30 seconds instead.

Fixes #294.

I was unable to reproduce the issue again with the 10 second wait (typical!), but I was able to repro by setting the timeout *lower*.

Thanks!

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -31,6 +31,8 @@ CONFIG_DIR = Path(click.get_app_dir("catt"))
 CONFIG_PATH = Path(CONFIG_DIR, "catt.cfg")
 STATE_PATH = Path(CONFIG_DIR, "state.json")
 
+WAIT_PLAY_TIMEOUT = 30
+
 
 class CattTimeParamType(click.ParamType):
     def convert(self, value, param, ctx):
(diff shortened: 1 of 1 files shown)


skorokithakis (approved):

skorokithakis:
Looks good, thanks! I edited the title so the changelog credits you.

PR #304, opened 2020-10-15
Rework verification of cli option device

### LINKED ISSUE #253: set_default improvements
While trying to test the directory feature from #251, I found some issues with `set_default`:

* `write_config` is deprecated, correct? If so, its help text should read "DEPRECATED:" at the start.
* `set_default` lets you add a device name or IP as the default, but we never show the user the names and IPs of devices, do we? We should make that clearer, or make `set_default` print a list of devices if no device is specified.

theychx:
fixes #253

The problem with the previous approach to getting config data into `click`, was that once the device info reached the subcmd, there was no way for it to tell whether it came from the cli or the config.
With this new approach we can do just that, and so I`ve made it mandatory to specify the device as a cli option, whenever the user wants to manipulate the config file. Not exactly what you asked for :smiley: , but I think that this is pretty clean UX.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -96,18 +96,20 @@ def process_subtitles(ctx, param, value):
     return value
 
 
-def process_device(ctx, param, value):
+def process_device(device_desc, aliases):
     """
     Resolve real device name when value is an alias.
 
-    :param value: Can be an ip-address or a name (alias or real name).
-    :type value: str
+    :param device_desc: Can be an ip-address or a name (alias or real name).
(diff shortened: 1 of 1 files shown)


skorokithakis:
Thanks for this, it looks good to me! I think we might want to refactor the device into its own type at some point, so `process_device()` can return a `Device`, which could have a `from_cli` attribute.

This looks good regardless, though, thanks again!

skorokithakis (approved):

PR #306, opened 2020-10-18
Improve setup_cast

theychx:
I've moved the inactivity check to setup_cast where it belongs, and also made setup_cast less clunky.

### DIFF
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -83,22 +83,18 @@ def setup_cast(device_desc, video_url=None, controller=None, ytdl_options=None, 
 
     if controller:
         app = get_app(controller, cast_type, strict=True)
-    elif prep == "app":
-        if stream:
-            if stream.is_local_file:
-                app = get_app("default")
-            else:
-                app = get_app(stream.extractor, cast_type, show_warning=True)
-        else:
-            app = get_app("default")
-    else:
+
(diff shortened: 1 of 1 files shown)


catt/controllers.py
```diff
@@ -211,10 +209,7 @@ def __init__(self, current_state, states, invert=False, fail=False):
         self._state_event = threading.Event()
         self._current_state = current_state
         if self._current_state in self._states_waited_for:
-            if fail:
-                raise ListenerError("Condition is already met (fail is set)")
```
theychx, line 215: (unrelated to the PR)
I've removed this strange bit of code, as it isn't used anywhere. I honestly don't remember why I put it there.

skorokithakis: I feel you, I hate bits of code whose purpose I don't remember 😛 That's why I usually leave comments anywhere that's not completely obvious.

theychx: I'm pretty sure it's never been used. I must have convinced myself that it would be useful at some point.

theychx:
So, uhm, what do you thnik of the PR? :smiley:

skorokithakis:
I quite like it! Sorry, I didn't have time to review thoroughly until now, I just got to the PC. Approved!

skorokithakis (approved):

PR #307, opened 2020-10-22
Cast 'n' seek

### LINKED ISSUE #286: Feature request: cast with seek
Feature request: a new `--seek` flag for `catt cast`, so that initialization of the cast, and setting the seek position to start playing from, can be done in a single command.

theychx:
fixes #286 :horse_racing: :horse_racing: :horse_racing:

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -160,7 +160,7 @@ def cli(ctx, device):
     help="YouTube-DL option. "
     "Should be passed as `-y option=value`, and can be specified multiple times (implies --force-default).",
 )
-@click.option("--seek", type=CATT_TIME, metavar="TIME", help="Start playback at specific timestamp.")
+@click.option("-t", "--seek-to", type=CATT_TIME, metavar="TIME", help="Start playback at specific timestamp.")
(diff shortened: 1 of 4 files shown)


skorokithakis:
You're on fire! I will review when I'm at the PC and get back to you.

catt/cli.py
```diff
@@ -160,6 +160,7 @@ def cli(ctx, device):
     help="YouTube-DL option. "
     "Should be passed as `-y option=value`, and can be specified multiple times (implies --force-default).",
 )
+@click.option("--seek", type=CATT_TIME, metavar="TIME", help="Start playback at specific timestamp.")
```
skorokithakis, line 163: Could we rename this to "seek-to"? I think that's a bit more explicit as a name, what do you think?

skorokithakis (approved):
Looks almost ready, I only have one comment!

catt/cli.py
theychx: You're right. Fixed.

skorokithakis (approved):
Approved even more!

PR #309, opened 2020-11-01
Don't do discovery dance when checking ip validity

theychx:
:man_dancing: :woman_dancing: :man_dancing: :woman_dancing: :man_dancing: :woman_dancing: :man_dancing: :woman_dancing:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 
-## Unreleased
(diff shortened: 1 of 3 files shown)


skorokithakis (approved):
Looks good to me!

skorokithakis:
Hmm, can you add `fix: ` to the start of your commit so you get changelog credit?

PR #310, opened 2020-11-28
Add catt/__main__.py to enable 'python -m catt'

sethmlarson:
Thanks for the tool, used it for the first time minutes ago and worked flawlessly.

Immediately had the suggestion to allow `$ python -m catt ...` in case the entry point isn't installed properly (ie pyenv)

### DIFF
--- a/catt/__main__.py
+++ b/catt/__main__.py
@@ -1,0 +1,2 @@
+from .cli import main
+main()


skorokithakis:
Thank you, I think this is good. Tests are breaking for unrelated reasons, but I think this can be merged.

skorokithakis:
Can you run pre-commit on the files before I merge?

PR #313, opened 2020-12-23
fix docker example

m1k1o:
Given docker example would not work, because container is put in its own network, where are no devices. Forcing docker to join host network allows it to see devices. Also `--rm` added, so that example will be deleted as soon as it is executed.


skorokithakis (approved):
You are correct, thanks!

PR #345, opened 2021-05-19
Allow use of info cmds on idle/inactive cc

### LINKED ISSUE #343: When Google Hub is inactive, I cannot get volume status
Volume of Google Homes, are regularly turned down, goes automatically and / or is done by the children.
Now I wanted to have a script check that volume is not 100%, that script sets the relevant Google Home volume to 100%.

But when I give command: catt -d "Living room" status
Then I get: Error: Chromecast is inactive.

I would just like to see the volume, to be able to perform further action with this by means of a script.

theychx:
Fixes #343 

:loudspeaker: :loud_sound: :loudspeaker: :loud_sound: :loudspeaker: :loud_sound:


skorokithakis (approved):
Looks good, thank you!

PR #369, opened 2021-12-16
Fixed leftover youtube-dl references.

anthonyrocom:


catt/cli.py
```diff
@@ -180,10 +180,11 @@ def cli(ctx, device):
 @click.option(
     "-y",
     "--ytdl-option",
+    "--ytdlp-option",
```
skorokithakis, line 183: I don't think this needs to be changed, and it would break a bunch of stuff. It can stay as it was, I think.

anthonyrocom: Agreed

skorokithakis: Can you change it back though?

skorokithakis:
```suggestion
    "--ytdl-option",
```

skorokithakis:
@anthonyrocom you've tested this and it works properly, right?

anthonyrocom:
> @anthonyrocom you've tested this and it works properly, right?

Is YouTube working? Yes it's working and I tested that.

skorokithakis (approved):

catt/cli.py
```diff
@@ -180,10 +180,11 @@ def cli(ctx, device):
 @click.option(
     "-y",
     "--ytdl-option",
+    "--ytdl-option",
```
skorokithakis, line 183:
```suggestion
```

skorokithakis (approved):

PR #382, opened 2022-01-25
Catching the ImportError on systems running Python versions older than 3.8

bronikowski:
The importlib.metadata used to live as importlib_metadata, this try…except will make catt runnable on older installations


skorokithakis (approved):
Yep, this should work great, thanks!

PR #392, opened 2022-04-07
refactor: Pass extra args to the media controller

l1l0l:
This will allow the possibility to init the media controller with a title, current_time and a thumb


skorokithakis (approved):
Looks good, thanks!

Issue #376, opened 2022-01-14
Casting a YouTube playlist fails with an AttributeError


alexdelorenzo:
Thanks for releasing `catt`, it's a nice tool. I had an issue with its most recent release on PyPI.

Running this:
```bash
$ catt -d my_device cast "https://www.youtube.com/playlist?list=PLjHf9jaFs8XVAQpJLdNNyA8tzhXzhpZHu"
```
Fails with this exception:
```python
Traceback (most recent call last): 
  File "~/.pyenv/versions/3.10.1/bin/catt", line 8, in <module> 
    sys.exit(main())
<snip>
  File "~/.pyenv/versions/3.10.1/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "~/.pyenv/versions/3.10.1/lib/python3.10/site-packages/catt/cli.py", line 243, in cast
    if not random_play and cst.playlist_capability and stream.playlist_all_ids:
  File "~/.pyenv/versions/3.10.1/lib/python3.10/site-packages/catt/stream_info.py", line 143, in playlist_all_ids
    if self.is_playlist and self._entries and self._entries[0].get("id"):
AttributeError: 'str' object has no attribute 'get'. Did you mean: 'playlist_id'?
```
Versions:
```bash
$ python --version
Python 3.10.1

$ catt --version
catt v0.12.5, Zaniest Zapper.
```

This error might be related to https://github.com/skorokithakis/catt/issues/367.

skorokithakis:
Hm, yes, unfortunately I think this is related to #367, and so will take a bit of time to fix. I will see if I have some time to work on this, but PRs here are definitely appreciated.

skorokithakis:
I will close this as a duplicate for now, please discuss in #367.

PR #401, opened 2022-06-22
Fix youtube playlist issues

### LINKED ISSUE #367: Fix yt-dlp support
Actually, it seems fe9097565767e255b146b5ad35f5a7486c5fa922 needs a fixup to port `stream_info.py` (and also `test_catt.py`) to use `yt-dlp`.

Here are the docs for that: https://github.com/yt-dlp/yt-dlp#embedding-yt-dlp, and this is the most relevant bit:

> **Tip**: If you are porting your code from youtube-dl to yt-dlp, one important point to look out for is that we do not guarantee the return value of `YoutubeDL.extract_info` to be json serializable, or even be a dictionary. It will be dictionary-like, but if you want to ensure it is a serializable dictionary, pass it through `YoutubeDL.sanitize_info` as shown in the example above

_Originally posted by @FranciscoPombal in https://github.com/skorokithakis/catt/issues/354#issuecomment-991135771_

theychx:
Originally reported in #376.
Casting of direct media urls seem to work again, so I updated the `yt-dlp` requirement.

fixes #367

### DIFF
--- a/catt/stream_info.py
+++ b/catt/stream_info.py
@@ -166,7 +166,7 @@ class StreamInfo:
 
     def _get_stream_preinfo(self, video_url):
         try:
-            return self._ydl.sanitize_info(self._ydl.extract_info(video_url, process=False))
+            return self._ydl.extract_info(video_url, process=False)
         except yt_dlp.utils.DownloadError:
             # We sometimes get CI failures when testing with YouTube videos,
             # as YouTube throttles our connections intermittently. We evaluated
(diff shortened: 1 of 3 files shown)


catt/stream_info.py
```diff
@@ -166,7 +166,7 @@ def set_playlist_entry(self, number):
 
     def _get_stream_preinfo(self, video_url):
         try:
-            return self._ydl.sanitize_info(self._ydl.extract_info(video_url, process=False))
```
theychx, line 169: `sanitize_info` stringifies complex object references, which breaks youtube playlists (which are generators).

PR #403, opened 2022-06-23
fix: Rework broken get_cast_with_ip

### LINKED ISSUE #402: `catt==0.12.8` with `PyChromecast==12.1.4` fails with a `TypeError`
Here's the example using the latest `catt` with the latest `PyChromecast`:
```bash
$ catt -d $(dig +short chromecast.lan) cast "https://www.youtube.com/watch?v=YzuvMHvQiiA"
```
 ```python3
  File "~/.local/lib/python3.9/site-packages/catt/cli.py", line 232, in cast                                       
    cst, stream = setup_cast( 
  File "~/.local/lib/python3.9/site-packages/catt/controllers.py", line 78, in setup_cast
    cast = get_cast(device_desc) 
  File "~/.local/lib/python3.9/site-packages/catt/discovery.py", line 111, in get_cast                                                  
    cast = get_cast_with_ip(cast_desc, DEFAULT_PORT)                    
  File "~/.local/lib/python3.9/site-packages/catt/discovery.py", line 79, in get_cast_with_ip
    cast = pychromecast.Chromecast(cast_ip, port=port, tries=1)
TypeError: __init__() got an unexpected keyword argument 'port'
```
After the 9.x releases of PyChromecast, `pychromecast.Chromecast` doesn't take a `port` argument anymore. 

Downgrading PyChromecast fails with Catt, because newer versions of Catt use PyChromecast's updated API elsewhere.

The last Catt version I got working is `catt==0.12.2` with `PyChromecast==9.2.0`. Versions of Catt above 0.12.2 that depend on newer PyChromecast versions would fail, as well.

As an aside, I have a couple of projects built against PyChromecast, and while it is an excellent library, in my experience, semantic versioning isn't always followed by the project. 
I found the need to pin PyChromecast to major release versions, and sometimes to specific point releases, because the project will sometimes make breaking API changes in minor version bumps.

### PR DESCRIPTION
**Pull request by theychx**
Some `pychromecast` changes I missed...
Given that specifying a device with an ip-address is broken in `0.12.8`, you should make a new release again. Sorry about that.

fixes #402

### DIFF
--- a/catt/discovery.py
+++ b/catt/discovery.py
@@ -40,8 +40,8 @@ def get_cast_infos() -> List[pychromecast.CastInfo]:
     """
     Discover all available devices, and collect info from them.
 
-    :returns: Various device info, packed in CastInfo namedtuple.
-    :rtype: pychromecast.CastInfo
+    :returns: List of CastInfo namedtuples.
+    :rtype: List[pychromecast.CastInfo]
     """
 
     return [c.cast_info for c in get_casts()]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: catt/discovery.py [hunk 1]
```diff
@@ -40,8 +40,8 @@ def get_cast_infos() -> List[pychromecast.CastInfo]:
     """
     Discover all available devices, and collect info from them.
 
-    :returns: Various device info, packed in CastInfo namedtuple.
-    :rtype: pychromecast.CastInfo
+    :returns: List of CastInfo namedtuples.
+    :rtype: List[pychromecast.CastInfo]
```
skorokithakis, line 44: Since we already have this information in the signature, does it make sense to duplicate it here?

skorokithakis (approved):

catt/discovery.py
skorokithakis: That'd be good, they were helpful when they were the only ones, but now they're just something that can go out of date.

skorokithakis:
Sorry, released 0.12.9!

PR #406, opened 2022-06-28
Fix discovery function

soreau:
The path to get the ip from pychromecast is now c.socket_client.host


skorokithakis (approved):

skorokithakis:
Looks good, thanks!

PR #410, opened 2022-07-31
Fix the 10 minutes timeout issue

swiergot:
I have forked DashCast to implement a fix for the timeout problem:

https://github.com/swiergot/dashcast

It is registered at Google under ID 5C517DAC.

This pull request a) switches to this new DashCast version and b) adds a new option to cast_site to enable the fix (for backward compatibility the old behaviour remains the default).

Please note that due to security constraints the new method requires the casted website to be served over HTTPS.

### DIFF
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -315,16 +315,22 @@ def cast(
 
 @cli.command("cast_site", short_help="Cast any website to a Chromecast.")
 @click.argument("url", callback=process_url)
+@click.option(
+    "-i",
+    "--iframe",
+    is_flag=True,
+    help="Load website into an iframe to avoid timeout (requires HTTPS).",
+)
 @click.pass_obj
-def cast_site(settings, url):
+def cast_site(settings, url, iframe):
     cst = setup_cast(
         settings["selected_device"],
         controller="dashcast",
         action="load_url",
         prep="app",
     )
-    click.echo('Casting {} on "{}"...'.format(url, cst.cc_name))
-    cst.load_url(url)
+    click.echo('Casting {} on "{}" (iframe={})...'.format(url, cst.cc_name, iframe))
+    cst.load_url(url, force=not iframe)
 
 
 @cli.command(short_help="Add a video to the queue (YouTube only).")
--- a/catt/controllers.py
+++ b/catt/controllers.py
@@ -7,7 +7,6 @@ from typing import Optional
 
 import pychromecast
 from pychromecast.config import APP_BACKDROP as BACKDROP_APP_ID
-from pychromecast.config import APP_DASHCAST as DASHCAST_APP_ID
 from pychromecast.config import APP_MEDIA_RECEIVER as MEDIA_RECEIVER_APP_ID
 from pychromecast.config import APP_YOUTUBE as YOUTUBE_APP_ID
 from pychromecast.controllers.dashcast import DashCastController as PyChromecastDashCastController
@@ -25,6 +24,9 @@ from .util import echo_warning
 GOOGLE_MEDIA_NAMESPACE = "urn:x-cast:com.google.cast.media"
 VALID_STATE_EVENTS = ["UNKNOWN", "IDLE", "BUFFERING", "PLAYING", "PAUSED"]
 CLOUD_APP_ID = "38579375"
+
+DASHCAST_NAMESPACE = "urn:x-cast:es.offd.dashcast"
+DASHCAST_APP_ID = "5C517DAC"
 
 
 class App:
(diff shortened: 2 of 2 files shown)


skorokithakis:
Thank you! I'm a bit wary of relying on a third party for part of CATT's functionality, would it be possible to add that to CATT?

AFAIK we already have a custom receiver that's registered, it's on this branch:

https://github.com/skorokithakis/catt/blob/feat/custom-receiver/catt_receiver/receiver.html

catt/controllers.py
```diff
@@ -26,6 +25,9 @@
 VALID_STATE_EVENTS = ["UNKNOWN", "IDLE", "BUFFERING", "PLAYING", "PAUSED"]
 CLOUD_APP_ID = "38579375"
 
+DASHCAST_NAMESPACE = "urn:x-cast:es.offd.dashcast"
+DASHCAST_APP_ID = "5C517DAC"
```
skorokithakis, line 29:
```suggestion
DASHCAST_APP_ID = "1B3FF1BC"
```

skorokithakis:
I've deployed your fork to our custom receiver, would it be possible to deploy your code upstream for inclusion?

theychx:
@swiergot how is your approach different from the iframe approach already present in dashcast?
As I recall, using an iframe means a lot of sites won't work.

skorokithakis:
@theychx I'm glad you showed up, I was out of my depth here :P

swiergot:
> I've deployed your fork to our custom receiver, would it be possible to deploy your code upstream for inclusion?

Where exactly did you deploy it? The link you gave returns 404.

What upstream do you have in mind? https://github.com/stestagg/dashcast?

> @swiergot how is your approach different from the iframe approach already present in dashcast? As I recall, using an iframe means a lot of sites won't work.

So now the options are:

1. DashCast in window.location mode (old or new)

    - works with all websites,
    - times out after 10 minutes (unless the target website is a receiver itself - that's what [ha-catt-fix](https://github.com/swiergot/ha-catt-fix) does).

2. Old DashCast in iframe mode

    - requires the target URL to be HTTPS,
    - doesn't work with websites that block iframe,
    - **times out** after 10 minutes.

3. New DashCast in iframe mode

    - requires the target URL to be HTTPS,
    - doesn't work with websites that block iframe,
    - **does NOT time out** after 10 minutes.

The third, new option will be a huge improvement for many people.

skorokithakis:
> Where exactly did you deploy it? The link you gave returns 404.

It's at https://catt.stavros.io/, the link 404s because I moved the file when I added your fork to it.

> What upstream do you have in mind? https://github.com/stestagg/dashcast?

Yes, though I now see it hasn't been updated in 6 years, so it'd probably be pointless.

> The third, new option will be a huge improvement for many people.

I agree that having the option is worthwhile, even if it doesn't work for everyone. Since it's an option, they can choose "works with every site for 10 minutes" or "works with some sites for however long".

theychx:
@swiergot OK, thanks for clarifying.

skorokithakis:
@theychx does this look good to you?

swiergot:
> It's at https://catt.stavros.io/, the link 404s because I moved the file when I added your fork to it.

Ok, that looks fine.

> Yes, though I now see it hasn't been updated in 6 years, so it'd probably be pointless.

Well, I've done it anyway, won't hurt (https://github.com/stestagg/dashcast/pull/21).

> I agree that having the option is worthwhile, even if it doesn't work for everyone. Since it's an option, they can choose "works with every site for 10 minutes" or "works with some sites for however long".

It will work for people casting Home Assistant and similar.

Although after playing with it today I can see some weird Home Assistant UI issues (eg. mini-graph-card not showing the graph). I did not observe that with the ha-catt-fix method.

catt/cli.py
```diff
@@ -315,16 +315,22 @@ def cast(
 
 @cli.command("cast_site", short_help="Cast any website to a Chromecast.")
 @click.argument("url", callback=process_url)
+@click.option(
+    "-i",
+    "--iframe",
+    is_flag=True,
+    help="Load website into an iframe to avoid timeout (requires HTTPS).",
```
skorokithakis, line 322:
```suggestion
    help="Load website into an iframe to avoid timeout (requires HTTPS, and the site must allow loading into an iframe).",
```

skorokithakis (changes requested):

theychx:
@skorokithakis I'm not much of a javascript expert, so cant give any immediate critique.

@swiergot Using `cast_site` repeatedly doesn't appear to be working
(in either mode using https://ard.de and https://bbc.co.uk)

swiergot:
> @swiergot Using cast_site repeatedly doesn't appear to be working

Thanks, I will be on vacation for the next two weeks, will look into it afterwards.

skorokithakis:
Thanks, enjoy your vacation!

PR #414, opened 2022-09-10
doc: Document macOS config path in README

mk12:
This updates the README to document the config path as ~/Library/Application Support/catt/catt.cfg. It does not read from ~/.config/catt/catt.cfg on macOS.


skorokithakis (approved):

skorokithakis:
Thank you!

PR #427, opened 2023-04-23
Add volumemute command

neurodiv-eric:
With this addition, CATT should now have the ability to toggle between mute states while automatically preserving the previous volume level. This function will also prevent a volume preview tone on some devices.

Example: `catt -d <DEVICE_OR_IP> volumemute <true | false>`
Note that the default value is `true`.

### Does this work for all devices?
_Not exactly._ I also have a Lenovo Smart Display. While that device mutes and unmutes as expected, it unfortunately still emits a volume tone.

I only have the two device models, so feedback on which brands are/are not working would be great. I suspect it to be the same as Home Assistant's `media_player.volume_mute` service.

### DIFF
--- a/catt/api.py
+++ b/catt/api.py
@@ -162,7 +162,7 @@ class CattDevice:
 
     def volumemute(self, muted: bool) -> None:
         """
-        Enable mute on supported devices. Added by @neurodiv-eric
+        Enable mute on supported devices.
 
         :param muted: Whether to mute the device. (valid values: true or false).
         """
--- a/catt/api.py
+++ b/catt/api.py
@@ -160,6 +160,15 @@ class CattDevice:
 
         self.controller.volumedown(delta)
 
+    def volumemute(self, muted: bool) -> None:
+        """
+        Enable mute on supported devices. Added by @neurodiv-eric
+
+        :param muted: Whether to mute the device. (valid values: true or false).
+        """
+
+        self.controller.volumemute(muted)
+
 
 def discover() -> List[CattDevice]:
     """Perform discovery of devices present on local network, and return result."""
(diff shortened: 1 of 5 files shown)


skorokithakis:
Thank you! I'll review shortly.

catt/api.py
```diff
@@ -160,6 +160,15 @@ def volumedown(self, delta: float) -> None:
 
         self.controller.volumedown(delta)
 
+    def volumemute(self, muted: bool) -> None:
+        """
+        Enable mute on supported devices. Added by @neurodiv-eric
```
skorokithakis, line 165: Could you move the credit here to the CONTRIBUTING file? Docstrings should only be for function documentation. The changelog will also show your name automatically.

neurodiv-eric:
Removed attribution in the docstring- anything else look off?

catt/api.py
neurodiv-eric: Done! Anything else look off?

catt/util.py
```diff
@@ -38,6 +38,9 @@ def echo_status(status):
     if status.get("volume_level"):
         click.echo("Volume: {}".format(status["volume_level"]))
 
+    if status.get("volume_muted"):
+        click.echo("Volume Muted: {}".format(status["is_volume_muted"]))
```
skorokithakis, line 42:
```suggestion
        click.echo("Volume muted: {}".format(status["is_volume_muted"]))
```

skorokithakis (approved):
Looks good to me, just one tiny text consistency correction. Also, why is it called `volumemute` and not just `mute`?

catt/util.py
neurodiv-eric: I chose `volumemute` because the attribute from pychromecast data is `volume_muted`. It just seemed more consistent. I'm not particularly attached to it, so changing it to `mute` would also be fine.

skorokithakis (approved):

PR #433, opened 2023-09-19
fix: set explicit build-backend

mweinelt:
The project seems to use poetry to manage its dependencies, and defining the build-system allows building it using the generic `build` frontend, which distros heavily rely on.

See https://github.com/python-poetry/poetry-core#usage.


skorokithakis (approved):

skorokithakis:
Thank you!

PR #455, opened 2025-01-27
Migrate to `uv`

seguri:


skorokithakis:
This looks good, thank you. I want to run it on CI to make sure there are no errors, but the CI seems to have failed. I'll try to fix it shortly and report back.

seguri:
Great. As I said, double check what I wrote, as it is partially done by AI, as I've never seen `uv` before. For example, I discovered in the meantime that a simpler way to setup a dev environment is just `uv sync` (creates venv + installs dev dependencies if pyproject.toml is properly setup).

PR #462, opened 2025-04-23
REF: controllers.py: Update self._cast.status._asdict() to .__dict__ …

westurner:
…so that `catt info` works


skorokithakis (approved):

skorokithakis:
Thanks!

westurner:
Thanks!

PR #467, opened 2025-08-04
More control over subtitles

### LINKED ISSUE #117: catt doesn't allow to select subtitles inside the container
The `--subtitle` option looks for either an URL or a local file. There is no way to specify a track inside the container.
`catt --info` does show that the subtitle tracks present inside the container.
It would be nice if those could be selected.

Not sure how much work that would be. And in any case, thank you for `catt`. It's so much better than going through chromium or an app on my phone!

### LINKED ISSUE #412: font size subtitles
Would it be possible to set the font size for subtitles yourself?

medape:
Hello again! These commits add some conveniences to toggle subtitles:

## Allow the use of TTML subtitles
Because why not? Even my 1st-gen ChromeCast supports them... partially.

Like VTT; only positioning, bold, italics and underline are honoured. _Unlike_ VTT, some HTML codes cause the receiver to stop interpreting subtitles at all. 

## Automatically find subtitles on YouTube (default app only)
When using the `catt cast -f URL` (`catt cast --force-default URL`) command, all available subtitles are automatically found, and the best options (VTT > TTML > SRT) sent to the ChromeCast through the pychromecast parameter `media_info`.

- This extends to any platform that `yt-dlp` is compatible with.
- In the future, it should be possible to use the same mechanism to fix #412
- If a subtitle URL is provided with the flag `-s` of `--subtitles`, only that subtitle file is loaded. This is the undocumented behaviour of the pychromecast module when both arguments `subtitles` and `media_info` are used, and I didn't find it prudent to work around this for now.

## Add a new menu to toggle subtitles on and off
The menu is quite basic, and TRACK_ID is always a number:
```
Usage: catt subs [OPTIONS] [TRACK_ID]

Options:
  -n, --off   Hides all subtitles
  -l, --list  Lists all subtitles
  -h, --help  Show this message and exit.
```

These work for the default app. The YouTube app uses a different controller, and I couldn't find the relevant documentation, so only `catt subs --off` is compatible with it.

## MIME type is only guessed for local files

Subtitle files in web servers are already served with MIME types, so we just read the MIME type from the HTTP response. This way, we don't need to check the URL, _and_ we can use URLs that don't end with .srt, .vtt or .ttml (like YouTube's URLs)

## What I couldn't do

- I was unable to get [EIA-608](https://en.wikipedia.org/wiki/EIA-608) subtitles to work; it seems like those are only ever found embedded in the video, and `yt-dlp` [does not attempt to retrieve them](https://github.com/yt-dlp/yt-dlp/issues/2859)
- for the same reason, this does NOT fix #117
- use these in the official YouTube app, as mentioned above


catt/subs_info.py
```diff
@@ -4,32 +4,60 @@
...
     converting it to webvtt if needed, and then exposing a path to a tempfile
     holding the subtitles, ready to be served.
-    An url to the (expected to be) served file is also exposed. The supplied
+    An url to the (expected to be) served file is also exposed.
+
+    For local subtitle files, the url points to this computer. The variable
+    self.local_subs will be True; CATT will detect this and start a web
+    server so that ChromeCast can fetch it. Only devices in the same network
+    can access this file.
     local_ip and port params are used for this purpose.
```
skorokithakis, line 21: Is this forgotten from the previous docstring?

medape: I'm sorry I'm late. The answer's yes and no.

`self.local_subs` is still used, but now it's not directly provided by `self._read_subs()`. instead, `self._read_subs()` provides a MIME type only if the subtitles file was originally remote, AND the script uses the existence of the MIME type to infer if the file was remote or not.

_This_ behaviour isn't documented anywhere, though, so I'll update the docstring right away.

skorokithakis (approved):
Looks good, thanks!

Issue #486, opened 2025-10-01
Scan in json format doesnt work


AbdullahGheith:
Environment:

```
root@radxa-zero3:~# uname -a
Linux radxa-zero3 6.1.115-vendor-rk35xx #1 SMP Wed Sep 17 00:59:42 UTC 2025 aarch64 GNU/Linux

```

```
root@radxa-zero3:~# python3 --version
Python 3.11.2
```

```
root@radxa-zero3:~# catt --version
catt v0.13.1, Zaniest Zapper.
```

Error:
```
root@radxa-zero3:~# catt scan --json-output
Traceback (most recent call last):
  File "/root/.local/bin/catt", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/catt/cli.py", line 705, in main
    return cli(obj=get_config_as_dict())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/click/core.py", line 1462, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/click/core.py", line 1383, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/click/core.py", line 1850, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/click/core.py", line 1246, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/click/core.py", line 814, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/pipx/venvs/catt/lib/python3.11/site-packages/catt/cli.py", line 535, in scan
    echo_json({d.friendly_name: d._asdict() for d in devices})
[...]

PR #488, opened 2025-11-01
Various changes

medape:
Each commit includes a changelog.

I've been working on this code for days on my computer. None of the commits should be broken (the script should work normally no matter which commit do you pick) but I had to edit my commits several times. In particular, my original debugging functions were a nightmare to set up, and replacing them force me to amend all the following commits.

My original intention was to unblock YouTube videos again, but so far, I haven't had any luck.

Newish videos on YouTube (from the last two years?) currently can't be downloaded in a merged format. The only options are:
- using an M3U8 manifest to download video and audio segments
- casting a video-only stream
- or casting an audio-only stream.

The first option works out of the box for the Tears of Steel test video at [`lolitemaultes/m3u8-urls`](https://github.com/lolitemaultes/m3u8-urls) and for streams from RTVE (Spanish public TV), but when CATT selects a YouTube M3U8 manifest, it just shows the Cast icon over black background.

I couldn't find any way to root my ChromeCast, or figure out debugging. I tried casting video only then adding audio tracks, but not only does the ChromeCast not add them, it stops responding to CATT commands other than `catt stop`.

Subtitling code should probably go to `subs_info.py`, but right now moving more code around will take me several days and result in no extra functionality.

Any help is welcome.

EDIT:
- The mandatory checks on my computer didn't check for `api.py`... but I just made a tiny change in order not to break the API. (If input `device_desc` to `StreamInfo()` was optional, the dict `settings` replacing it must be optional too).
- Edited the message for one of the first commits because in order to say `Click` instead of @click (I'm sorry if I pinged them for no reason).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.9
+  rev: v0.14.2
   hooks:
     - id: ruff
       args: [ --fix ]
     - id: ruff-format
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v1.17.1
+  rev: v1.18.2
   hooks:
   - id: mypy
     name: Run MyPy typing checks.
--- a/catt/cli.py
+++ b/catt/cli.py
@@ -32,6 +32,7 @@ from .util import echo_status
 from .util import echo_warning
 from .util import hunt_subtitles
 from .util import is_ipaddress
+from .util import set_verbosity
 
 CONFIG_DIR = Path(click.get_app_dir("catt"))
 CONFIG_PATH = Path(CONFIG_DIR, "catt.cfg")
@@ -44,6 +45,8 @@ try:
     VERSION = version(PROGRAM_NAME)
 except Exception:
     VERSION = "0.0.0u"
+
+MAX_VERBOSITY = 4
 
 
 class CattTimeParamType(click.ParamType):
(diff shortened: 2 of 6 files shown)


skorokithakis:
Sorry, what does this PR do? "Various changes" isn't very descriptive, and each logically distinct change should really be in its own PR, so it can be merged independently. Thanks for your contribution!

medape:
I agree. Though, in this case, pretty much every commit here does its own thing.

Basically, this PR does four things:

- adds debugging functions, to help me and hopefully other people. Basically users can run `catt`, `catt -v`, `catt -vv` and `catt -vvv`. With each `v`, more info is output to the screen.
- add support to cast .m3u8 files from the computer (very limited for the moment; the current code only allows you to share one file, so these .m3u8 files can only point to online video and audio segments. Also a lot of formats fail, and Google's files fail for some reason)
- blacklist more videos from the older ChromeCasts (as seen in the commit message for 9f42f0f565a85040c331863662ed6a73a59d2b38) AND allow videos with unknown specs (useful when casting .m3u8 files with only one stream and no metadata)
- small quality-of-life features like fixing #486 and improving the output of `catt subs -l`

I can split it however you want me to, but all commits use the debugging functions I added, so I'd like you to review those first and tell me if you want them or not. If you want them, I'll split that into its own commit and wait for you to merge it; if no, I'll rewrite the other commits to not make use of them.

skorokithakis:
The verbosity levels seem like a great feature, though I feel like you're reinventing `logging`. Wouldn't the verbosity level just set the logging level `DEBUG/INFO/WARNING/ERROR` and use Python's built-in logging module? It does basically the exact same you're doing as well.

medape:
It is pretty much the same, I just thought `logging` had a lot of features we could do without, and required some setup to adapt to this program. Like, just using `logging` is nicer and more extensible, but this felt just as simple, or simpler in practical terms.

You're the boss, though. If you're okay with adding a new dependency for CATT, I can do a couple tests with the `logging` module and rewrite this.

skorokithakis:
Since `logging` is in the stdlib, I'd prefer to not reinvent the wheel and use the existing module for this. The added advantage is that every contributor will be familiar with it already, so there's less cognitive load with it too.

medape:
Roger that. I may take a bit, but I'll close this PR and open a new one only to add logging. Then I'll create a separate one for video selection and QOL, then a new one for M3U8.

PR #491, opened 2026-03-10
Title, Volume and remote Mp3 files fix

AbdullahGheith:
Added ability to provide title (useful for the mp3 media player) and volume when casting. 

Also remote mp3 files didnt work before, now they do instead of saying "no suitable format found"

Example:

`catt -d Office cast https://file-examples.com/storage/fede0b421769af5709b2497/2017/11/file_example_MP3_700KB.mp3 --title "This is a title" --volume 20`

oh, also output as json gives more info now

### DIFF
--- a/tests/test_catt.py
+++ b/tests/test_catt.py
@@ -58,6 +58,13 @@ class TestThings(unittest.TestCase):
         self.assertTrue(stream.is_remote_file)
         self.assertEqual(stream.extractor, "twitch")
 
+    def test_stream_info_direct_link(self):
+        url = "https://file-examples.com/storage/fede0b421769af5709b2497/2017/11/file_example_MP3_700KB.mp3"
+        stream = StreamInfo(url)
+        self.assertEqual(stream.video_url, url)
+        self.assertTrue(stream.is_remote_file)
+        self.assertTrue(stream._is_direct_link)
+
 
 if __name__ == "__main__":
     import sys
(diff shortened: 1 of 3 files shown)


tests/test_catt.py
```diff
@@ -58,6 +58,13 @@ def test_stream_info_other_video(self):
         self.assertTrue(stream.is_remote_file)
         self.assertEqual(stream.extractor, "twitch")
 
+    def test_stream_info_direct_link(self):
+        url = "https://file-examples.com/storage/fede0b421769af5709b2497/2017/11/file_example_MP3_700KB.mp3"
```
skorokithakis: Hm this URL doesn't exist, is it supposed to?

AbdullahGheith: Hmm... Annoying. I pulled the URL from here: https://file-examples.com/index.php/sample-audio-files/sample-mp3-download/

But seems like the URL is generated for the session only and not a direct link.

I will find another one

AbdullahGheith: I uploaded the mp3 to my own server instead.

skorokithakis (approved):
Hm, I'm not crazy about the tests depending on someone's server, but it's good enough for now, thanks!

AbdullahGheith:
I know its not the best solution, but i dont know how else we would test remote mp3 playback without it. If it stops working someday, we can also just remove the test.

I see that it failed the pre-commit because of formatting. I just formatted and pushed again.

skorokithakis (approved):

skorokithakis:
Makes sense, thanks!

